PR #2, opened 2024-02-14
Tighter FFI interop, metrics support, sdk-core updates, and numerous bugfixes

iand675:
There were some platform-specific memory access surprises that cropped up when doing CI testing in the Mercury backend code. It prompted me to move towards a more disciplined way to sync types (`cbindgen` on the Rust side, hsc2hs on the Haskell side). Additionally, there were some fun concurrency issues that were most easily fixed by migrating from `Ptr` usage to `ForeignPtr` usage throughout most of the Haskell code. A number of those uses are overkill, but they make it less likely that we'll shoot ourselves in the foot while trying to drive adoption. Once there's more testing in place, we can selectively move back to raw pointers to improve GC costs a smidge.


app/Main.hs
```diff
@@ -97,9 +97,6 @@ main = do
   hSetBuffering stdout NoBuffering
   forkIO $ do
     threadDelay 1000000
-    logs <- fetchLogs globalRuntime
```
iand675, line 100: There's no longer a `globalRuntime` since we have to initialize it with a metrics configuration, so these core SDK logs don't really make sense to use anymore. They are quite chatty, so I'll revisit how to integrate them better in a future PR.

core/rust/src/runtime.rs
```diff
@@ -144,57 +201,46 @@ impl Runtime {
...
+  pub level: String,
+  pub fields: HashMap<String, serde_json::Value>,
+  pub span_contexts: Vec<String>
 }
 
 #[no_mangle]
 pub extern "C" fn hs_temporal_runtime_fetch_logs(
   runtime: *mut RuntimeRef,
-) -> *const CArray<CCoreLog> {
+) -> *const CArray<CArray<u8>> {
```
iand675, line 224: These core logs were easier to turn into JSON to get across the FFI boundary. I need to implement better HashMap interop eventually.

core/rust/src/worker.rs
```diff
@@ -185,7 +185,7 @@ pub extern "C" fn hs_temporal_drop_worker(worker: *mut WorkerRef) -> () {
 pub extern "C" fn hs_temporal_new_worker(client: *mut client::ClientRef, config: *const CArray<u8>, result_slot: *mut*mut WorkerRef, error_slot: *mut*mut CWorkerError) -> () {
   let client_ref = unsafe { client.as_ref() }.expect("client is null");
   let config_json = unsafe { CArray::raw_borrow(config).unwrap() };
-  let config = serde_json::from_slice(&config_json.as_rust().unwrap()).map_err(|err| WorkerError {
+  let config = serde_json::from_slice(&config_json.as_rust().unwrap().clone()).map_err(|err| WorkerError {
```
iand675, line 188: `.clone()` introduced here since the memory on the Haskell side is only guaranteed to be live during this call. I don't trust the Rust code to not try to just use slice references inside of the parsed values.

core/rust/src/runtime.rs
```diff
@@ -17,38 +22,90 @@ pub(crate) struct Runtime {
...
-  let runtime = CoreRuntime::new(
+fn init_runtime(telemetry_config: TelemetryOptions, late_telemetry_options: HsTelemetryOptions, try_put_mvar: extern fn(capability: Capability, mvar: *mut MVar) -> ()) -> Box<RuntimeRef> {
+  let mut runtime = CoreRuntime::new(
     telemetry_config,
     tokio::runtime::Builder::new_multi_thread(),
-  );
+  ).unwrap();
+
+  let _guard = runtime.tokio_handle().enter();
+  let core_meter: Arc<dyn CoreMeter> = match late_telemetry_options {
```
iand675, line 32: Other languages also provide support for providing `CoreMeter` implementations that use callbacks into the wrapper language, but that's a problem for the future.

core/rust/src/runtime.rs
```diff
@@ -17,38 +22,90 @@ pub(crate) struct Runtime {
...
-    , try_put_mvar
-    )
-  )
+pub extern "C" fn hs_temporal_init_runtime(telemetry_opts: *const CArray<u8>, try_put_mvar: extern fn(Capability, *mut MVar) -> ()) -> *mut RuntimeRef {
+  let telemetry_opts = unsafe {
+    CArray::raw_borrow(telemetry_opts).unwrap().as_rust().unwrap().clone()
+  };
+  let telemetry_opts: HsTelemetryOptions = serde_json::from_slice(
+    telemetry_opts.as_slice()
+  ).expect("Failed to parse");
```
iand675, line 97: Future me wants this function to return `Either InitializationError Runtime` on the HS side, but I think it's okay to defer that work.

core/rust/src/runtime.rs
```diff
@@ -17,38 +22,90 @@ pub(crate) struct Runtime {
...
+
+  let early_options = TelemetryOptionsBuilder::default()
+          .logging(Logger::Forward {
+            filter: construct_filter_string(Level::INFO, Level::ERROR),
+          })
+          .attach_service_name(true)
+          // .metrics(core_meter)
```
iand675, line 104: OTel and Prometheus core_meter implementations require the Tokio runloop to be initialized, but that doesn't happen until after we call `init_runtime`, which is why we pass in `telemetry_opts` as a separate arg below and then do late binding onto the runtime.

core/src/Temporal/Core/CTypes.hsc
```diff
@@ -0,0 +1,306 @@
+#include "temporal_bridge.h"
+
```
iand675, line 2: This is our `hsc` file. It lets us use macros to calculate size, alignment, and field peek/poke offsets, among other things. This will make us more portable across architectures / less prone to error as compared to me previously manually determining all of this.

core/src/Temporal/Core/CTypes.hsc
```diff
@@ -0,0 +1,306 @@
...
+    x <- peekElemOff dataPtr i
+    f x
+
+cArrayToText :: CArray Word8 -> IO Text
+cArrayToText CArray{..} = do
+  let len = fromIntegral size
+  fromPtr dataPtr len
+
+instance Storable a => Storable (CArray a) where
+  sizeOf _ = #{size CArray_u8}
```
iand675, line 50: Since C doesn't have generics, we have to use a concrete version of `CArray` here, but I'm pretty convinced that we'll never have to deal with pointer sizes changing between the types that it wraps.

core/src/Temporal/Internal/FFI.hs
```diff
@@ -79,94 +41,54 @@ withCArrayText :: Text -> (Ptr (CArray Word8) -> IO b) -> IO b
...
 -- intervening, because we don't want to drop things like `Client` while they're still in use.
 -- So we should return ForeignPtrs for things that need to stay alive, and then we can drop when we're done.
-makeTokioAsyncCall :: (ManagedRustValue e, RustRef e ~ Ptr e, ManagedRustValue a, RustRef a ~ Ptr a) 
-  => TokioCall e a 
-  -> (RustRef e -> IO f)
-  -> (RustRef a -> IO b)
-  -> IO (Either f b)
-makeTokioAsyncCall call readErr readSuccess = mask_ $ do
+makeTokioAsyncCall 
+  :: TokioCall err res
```
iand675, line 66: I tweaked this pretty heavily to be less fancy. We just chuck things into `ForeignPtr` values with optional finalizer callbacks instead of doing all of the `ManagedRustValue` stuff. That was where a lot of the memory management issues were creeping in.

src/Temporal/Bundle.hs
```diff
@@ -473,26 +465,27 @@ type WorkerConfigs env rec = rec (ConstFn (WorkerConfig env))
 --
 -- This function starts each worker concurrently, waits for them to initialize, and then returns
 -- a worker for each task queue.
-startTaskQueues :: forall rec m env. (TraversableRec rec, MonadLoggerIO m, MonadUnliftIO m) => Client -> WorkerConfigs env rec -> m (Workers rec)
+startTaskQueues :: forall rec m env. (TraversableRec rec, MonadUnliftIO m) => Client -> WorkerConfigs env rec -> m (Workers rec)
 startTaskQueues client conf = startWorkers conf >>= awaitWorkersStart
   where
     startWorkers :: rec (ConstFn (WorkerConfig env)) -> m (rec (ConstFn (Async Worker)))
-    startWorkers = Rec.traverse (\_ workerConfig -> async (startWorker client workerConfig))
+    startWorkers = Rec.traverse $ \_ workerConfig -> flip runLoggingT workerConfig.logger $ do
```
iand675, line 472: Logging was moved into the `WorkerConfig` because I'd eventually like to support different log levels for the worker implementation as compared to the `Workflow` and `Activity` implementations.

src/Temporal/Duration.hs
```diff
@@ -51,9 +51,9 @@ parseDuration t
 
 -- | Add two durations together. Durations are subject to integer overflow.
 addDurations :: Duration -> Duration -> Duration
-addDurations (Duration s1 ns1) (Duration s2 ns2) = Duration (s1 + s2 + fromIntegral s) $ fromIntegral ns
+addDurations (Duration s1 ns1) (Duration s2 ns2) = Duration (s1 + s2 + s) $ fromIntegral ns
     where
-      (s, ns) = (ns1 + ns2) `divMod` 1_000_000_000
+      (s :: Int64, ns :: Int64) = fromIntegral (ns1 + ns2) `divMod` 1_000_000_000
```
iand675, line 56: Fix subtle integer overflow opportunity.

src/Temporal/EphemeralServer.hs
```diff
@@ -39,19 +37,13 @@ data EphemeralServerError = EphemeralServerError ByteString
...
+withDevServer rt conf = bracket 
   (liftIO (startDevServer rt conf) >>= either (throwIO . EphemeralServerError) pure) 
   (\e -> liftIO (shutdownEphemeralServer e) >>= either (throwIO . EphemeralServerError) pure)
 
-withTestServer :: MonadUnliftIO m => TestServerConfig -> (EphemeralServer -> m a) -> m a
-withTestServer = withTestServer' globalRuntime
-
-withTestServer' :: MonadUnliftIO m => Runtime -> TestServerConfig -> (EphemeralServer -> m a) -> m a
-withTestServer' rt conf = bracket 
+withTestServer :: MonadUnliftIO m => Runtime -> TestServerConfig -> (EphemeralServer -> m a) -> m a
```
iand675, line 45: Since `Runtime` values have to be initialized now, we don't need the two different versions of each of these.

iand675 (commented):
lots of changes here are also driven by building with `-Wall -Werror` during development.

andrewsmith (approved):
There is a lot here! I skimmed all of the diffs and honestly I don't have a lot of input given my shallow pass.

PR #38, opened 2024-10-31
Update core sdk to a newer version

iand675:
I needed some new functionality (ability to force the core worker to fail a workflow if it is nondeterministic) available in newer versions of the core SDK, so it was a good opportunity to do some upgrades.

It does introduce one other notable change, which is that we have to call an async `validate` function on the worker before we're allowed to start it up fully. It currently just checks that the remote namespace exists, AFAICT.

A few other new things we should support in subsequent PRs:

- `Worker::replace_client`: https://github.com/temporalio/sdk-core/pull/721
- Client API key in headers: https://github.com/temporalio/sdk-core/pull/699


core/rust/src/worker.rs
```diff
@@ -42,6 +44,8 @@ pub struct WorkerConfig {
   max_activities_per_second: Option<f64>,
   max_task_queue_activities_per_second: Option<f64>,
   graceful_shutdown_period_millis: u64,
+  nondeterminism_as_workflow_fail: bool,
+  nondeterminism_as_workflow_fail_for_types: Vec<String>,
```
iand675, line 48: These are the main things we need to support replay testing– we have to make sure that in replay mode, it doesn't retry nondeterminism failures.

core/rust/src/worker.rs
```diff
@@ -401,6 +437,34 @@ impl WorkerRef {
   }
 }
 
+#[no_mangle]
+pub extern "C" fn hs_temporal_validate_worker(
```
iand675, line 441: New SDK versions require calling a worker validation function before starting the event loops. The only error currently that gets "thrown" ensuring that the namespace being connected to exists, but there will probably be more validations in the future.

sdk/src/Temporal/Worker.hs
```diff
@@ -466,17 +479,21 @@ data Worker = forall ty.
   , workerActivityLoop :: InactiveForReplay ty (Async ())
   , workerCore :: Core.Worker ty
   , workerTracer :: Tracer
+  , workerEvictionEmitter :: TChan Workflow.EvictionWithRunID
```
iand675, line 482: This is the analogue of how the Typescript SDK implements this: whenever a workflow is evicted, the history replay logic looks at the evictions to see what the reason is for a `Workflow` instantiation being deleted from the dictionary of running workflows.

sdk/src/Temporal/Worker.hs
```diff
@@ -466,17 +479,21 @@ data Worker = forall ty.
...
 
 
 startReplayWorker :: (MonadUnliftIO m, MonadCatch m) => Runtime -> WorkerConfig actEnv -> m (Temporal.Worker.Worker, Core.HistoryPusher)
 startReplayWorker rt conf = provideCallStack $ runWorkerContext conf $ do
   $(logDebug) "Starting worker"
-  (workerCore, replay) <- either throwIO pure =<< liftIO (Core.newReplayWorker rt conf.coreConfig)
+  let coreConfig' = conf.coreConfig {Core.nondeterminismAsWorkflowFail = True}
+  (workerCore, replay) <- either throwIO pure =<< liftIO (Core.newReplayWorker rt coreConfig')
   $(logDebug) "Instantiated core"
+  workerEvictionEmitter <- newBroadcastTChanIO
```
iand675, line 492: `newBroadcastTChanIO` avoids memory leaks by not storing entries in the unbounded queue if there aren't any consumers.

sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -626,7 +626,17 @@ instance MonadReadStateVar Query where
...
 newtype InstanceM (a :: Type) = InstanceM {unInstanceM :: ReaderT WorkflowInstance IO a}
-  deriving newtype (Functor, Applicative, Monad, MonadIO, MonadReader WorkflowInstance, MonadUnliftIO)
+  deriving newtype
+    ( Functor
+    , Applicative
+    , Monad
+    , MonadIO
+    , MonadReader WorkflowInstance
+    , MonadUnliftIO
+    , Catch.MonadThrow
```
iand675, line 636: Drive-by addition of some instances that are trivial to support

sdk/src/Temporal/Workflow/Worker.hs
```diff
@@ -308,6 +316,7 @@ handleActivation activation = inSpan' "handleActivation" (defaultSpanArguments {
             join $ atomically $ do
               currentWorkflows <- readTVar worker.runningWorkflows
               writeTVar worker.runningWorkflows $ HashMap.delete runId_ currentWorkflows
+              writeTChan worker.workerEvictionEmitter EvictionWithRunID {runId = runId_, eviction = removeFromCache}
```
iand675, line 319: I love STM

PR #39, opened 2024-11-14
fix `millisecond` and `microsecond` functions

turnage:
Right now, calling `milliseconds 10` results in `Duration { seconds = 0, nanoseconds = 10 }`


matto-mercury (approved):
seems right :white_check_mark:

PR #40, opened 2024-11-19
Set default retry limit

turnage:
At the moment, activities retry infinitely. This has [surprised people](https://mercurytechnologies.slack.com/archives/C05SERU25H9/p1732052868882799). Most recent [example](https://cloud.temporal.io/namespaces/production.uoqel/workflows/one-off-tasks%2FPrepareRiskOrganizationFeatures%2F72729dbb-2c32-4b18-8e43-6a9ecbc780fd/54a5107a-1008-4ccf-a07f-b7fab93a5ec4/history).

The API provides other, smarter ways to manage retry loops than this pr adds, such as `ScheduleToClose` timeouts and workflow execution timeouts, but most users don't discover those for a while.

Setting a default finite retry limit will alert on activity retry failure by promoting it to workflow failure, which may trigger a user to understand the timeout options.


sdk/src/Temporal/Common.hs
```diff
@@ -199,6 +199,17 @@ instance ToJSON RetryPolicy
 instance FromJSON RetryPolicy
 
 
+defaultRetryPolicy :: RetryPolicy
+defaultRetryPolicy =
+  RetryPolicy
+    { initialInterval = seconds 1
+    , backoffCoefficient = 2
+    , maximumInterval = Nothing
+    , maximumAttempts = 10
```
turnage, line 208: 0 -> 10

sdk/src/Temporal/Workflow/Types.hs
```diff
@@ -114,7 +114,7 @@ defaultStartActivityOptions t =
     , timeout = toStartActivityTimeoutOption t
     , scheduleToStartTimeout = Nothing
     , heartbeatTimeout = Nothing
-    , retryPolicy = Nothing
+    , retryPolicy = Just defaultRetryPolicy
```
turnage, line 117: Nothing -> default 

This is the only one with a default (workflows can still infinitely loop), but I think that's reasonable? The workflows are pure.

turnage:
We're going to provide this internally and hlint-ban referencing this function from the SDK. [Context](https://mercurytechnologies.slack.com/archives/C04N6SQH1FT/p1732106314434559?thread_ts=1732061709.741249&cid=C04N6SQH1FT).

PR #47, opened 2025-01-06
auto-formatting & linting

jkachmar:
## description

my LSP integration was set up to auto-format with `rustfmt`; in the spirit of doing some spring cleaning, I ran through `clippy`'s errors & suggestions as well.

I'll add a few notes below, but the only thing I'm really unsure about is whether the `not_unsafe_ptr_arg_deref` lint handling here makes sense (details in f025784d32a65ac60552358d67f94b99fbea9233).

### DIFF
--- a/core/rust/src/client.rs
+++ b/core/rust/src/client.rs
@@ -1,28 +1,28 @@
-use temporal_client::{
-  ClientOptions, ClientOptionsBuilder, ClientOptionsBuilderError, ConfiguredClient,
-  RetryClient, RetryConfig, TemporalServiceClientWithMetrics, TlsConfig,
-};
-use tonic::{metadata::{MetadataKey, errors::InvalidMetadataValue}};
+use crate::runtime::{self, Capability, HsCallback, MVar};
 use ffi_convert::*;
-use crate::runtime::{self, Capability, MVar, HsCallback};
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::ffi::CStr;
+use std::str::{from_utf8_unchecked, FromStr};
 use std::time::Duration;
-use serde::{Deserialize, Serialize};
+use temporal_client::{
+    ClientOptions, ClientOptionsBuilder, ClientOptionsBuilderError, ConfiguredClient, RetryClient,
+    RetryConfig, TemporalServiceClientWithMetrics, TlsConfig,
+};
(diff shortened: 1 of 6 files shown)


core/rust/src/worker.rs
```diff
@@ -733,14 +799,28 @@ pub extern "C" fn hs_temporal_history_pusher_push_history(
...
     history_pusher.close()
 }
 
+// TODO: [publish-crate]
+/// # Safety
+///
+/// Haskell FFI bridge invariants.
+///
+/// The caller must ensure that the argument is a live pointer to a [`HistoryPusher`], typically from across the FFI
+/// boundary after having been constructed by [`hs_temporal_new_replay_worker`].
```
jkachmar, line 819: I tried to add some more detailed safety docs for the first couple of these, but a lot of them are mechanical & I don't know the library well enough to provide useful commentary just yet.

core/rust/src/worker.rs
```diff
@@ -705,8 +767,12 @@ impl HistoryPusher {
     }
 }
 
+// TODO: [publish-crate]
+/// # Safety
+///
+/// Haskell <-> Tokio FFI bridge invariants.
```
jkachmar, line 771: It felt useful to call out functions that are meant to be invoked with `makeTokioAsyncCall` on the Haskell-side of things.

core/rust/src/client.rs
```diff
@@ -150,7 +154,7 @@ impl From<&RpcCall> for TemporalCall {
                 rust_vec.unwrap().clone()
             },
             retry: rpc_call.retry,
-            metadata: convert_hashmap(rpc_call.metadata),
+            metadata: unsafe { convert_hashmap(rpc_call.metadata) },
```
jkachmar, line 154: I think this is the only `clippy` error that wasn't `not_unsafe_ptr_arg_deref`

core/rust/src/client.rs
```diff
@@ -213,8 +217,12 @@ pub fn connect_client(
         })
 }
 
+// TODO: [publish-crate]
```
jkachmar, line 217: there's probably a better label to hang this off of, but I figured these are the sorts of things that we'd want to resolve before publishing a crate (at the very latest).

iand675 (approved):

PR #54, opened 2025-02-05
add 'extra_libraries' to SDK cabal file & fix test suite

jkachmar:
## description

* test suite now consults `HS_TEMPORAL_SDK_LOCAL_TEST_SERVER` to see if it should spawn ephemeral servers or use an external dev server for testing
  * if the variable exists, a dev server was started (e.g. with `temporal server start-dev`) in the background
* `temporal-sdk-core` now explicitly lists `temporal_bridge` in its `extra-libraries` stanza
  * this means that `cabal2nix` will add it to the argument list, and it more accurately reflects the fact that this is required for the library to function
  * local development should be unchanged; the custom `Setup.hs` will look for the `external_lib` flag and default to building & sourcing `temporal_bridge` w/ `cargo` if it's set to `False`

### DIFF
--- a/nix/temporal-sdk-pkgs.nix
+++ b/nix/temporal-sdk-pkgs.nix
@@ -16,19 +16,29 @@
     stdenv
     ;
 in {
-  temporal-sdk-core = overrideCabal (_attrs: {
-    extraLibraries = [
-      temporal-bridge.lib
+  temporal-sdk-core = lib.pipe
+    (final.callCabal2nix "temporal-sdk-core" ../core {
+      temporal_bridge = temporal-bridge.lib;
+    })
+    [
+      (enableCabalFlag "external_lib")
+      (overrideCabal (_attrs: {
+        libraryFrameworkDepends = lib.optionals stdenv.isDarwin (with darwin.apple_sdk.frameworks; [
+          CoreFoundation
+          Security
+          SystemConfiguration
+        ]);
+      }))
     ];
-    libraryFrameworkDepends = lib.optionals stdenv.isDarwin (with darwin.apple_sdk.frameworks; [
-      CoreFoundation
-      Security
-      SystemConfiguration
-    ]);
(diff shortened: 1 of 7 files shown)


iand675 (approved):

nix/temporal-sdk-pkgs.nix
```diff
@@ -16,19 +16,29 @@
...
 in {
-  temporal-sdk-core = overrideCabal (_attrs: {
-    extraLibraries = [
-      temporal-bridge.lib
+  temporal-sdk-core = lib.pipe
+    (final.callCabal2nix "temporal-sdk-core" ../core {
+      temporal_bridge = temporal-bridge.lib;
+    })
+    [
+      (enableCabalFlag "external_lib")
```
jkachmar, line 24: if we default this to `True` then we can set it in the repo's `cabal.project` to `False` for local dev w/ `cabal build` & avoid having to change the flag in any of our Nix files

sdk/test/IntegrationSpec.hs
```diff
@@ -318,10 +320,9 @@ spec = do
     setup :: (TestEnv -> IO ()) -> IO ()
     setup go = do
       let withServer f = do
-            if True -- Use local server instead of ephemeral server. Flip to use ephemeral server.
-              then do
-                f 7233
-              else do
+            lookupEnv "HS_TEMPORAL_SDK_LOCAL_TEST_SERVER" >>= \case
+              Just v | v == "true" || v == "0" -> f 7233
```
jkachmar, line 324: feels like there should be a better way to check for env var truthiness tbqh

iand675: Personally I feel like... if the env var exists, then it is indicative of intent?

iand675: Side-note, I generally expect `/= "0"` to mean true?

jkachmar: normally yeah but I was going w/ POSIX semantics because it's an env var; I'll just check for existence & call it at that

PR #56, opened 2025-02-11
big Nix refactor

jkachmar:
## description

this refactors a bunch of the Nix code along the following lines:
* switches from `crate2nix` to `rustPlatform.buildRustPackage`
* refactors from `flake-utils` to `flake-parts`, so that flake outputs can be defined in terms of Nix modules
* introduces `outputs.overlays` & `outputs.haskellOverlays`
  * `outputs.overlays.temporal-bridge` provides the `temporal_bridge` Rust package
  * `outputs.overlays.development` provides the GHC package sets that we use for our own development environment & CI
  * `outputs.haskellOverlays.hs-temporal-sdk` provides **just** the packages we manage herein
  * `outputs.haskellOverlays.dependencies.default` provides package overrides for all GHC packages in our matrix
  * `outputs.haskellOverlays.dependencies.ghc910` provides package overrides for **just** the GHC 9.10 package set

### notes

there's some stuff here that isn't as clean as I would like for it to be, but I think it's probably better to have this reviewed so we can support GHC 9.10 (& newer nixpkgs package sets) more easily.

of note, I haven't figured out how to deal with the `protogen` script quite yet so that's broken...


nix/overlays/haskell/ghc910-deps.nix
```diff
@@ -0,0 +1,34 @@
...
+      url = "https://github.com/google/proto-lens/pull/494/commits/7694e28e3ebcfcbcf00ff3eae73949d069c4f21a.diff";
+      relative = "proto-lens-protoc";
+      # There's a patch conflict in the autogenerated hpack header.
+      excludes = [
+        "package.yaml"
+        "proto-lens-protoc.cabal"
+      ];
+      hash = "sha256-qpBh8XGwb30VjIb0VGrLhu/k5az6QrL8XioE9pxDtw4=";
+    }))
+  ];
```
jkachmar, line 33: this whole section is kind of frustrating but it looks like all of this is just straight-up broken upstream so there's no way to work around it.

we could maybe avoid the `doJailbreak` & `appendPatch` for `proto-lens` stuff by just fetching the latest source and building those package directly (since they've updated to support GHC 9.10 but not yet cut a new release).

jkachmar: spoke with @iand675; decided to push this off

core/rust/default.nix
```diff
@@ -0,0 +1,66 @@
...
+}:
+let
+  cargoToml = builtins.fromTOML (builtins.readFile ./Cargo.toml);
+  crateHashes = builtins.fromJSON (builtins.readFile ./crate-hashes.json);
+in
+rustPlatform.buildRustPackage {
+  pname = cargoToml.package.name;
+  version = cargoToml.package.version;
+  # TODO: Apply some sort of Rust source filter to clean this.
+  src = ./.;
```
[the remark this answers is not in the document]
jkachmar: spoke with @iand675 & decided to push this till later

.envrc
```diff
@@ -1,5 +1,5 @@
 source_env_if_exists ~/.cargo/env
 
-use flake . --no-pure-eval
+use flake . --no-pure-eval --accept-flake-config
```
jkachmar, line 3: maybe this shouldn't be in the `.envrc`?

nix/overlays/haskell/deps.nix
```diff
@@ -0,0 +1,9 @@
+{ haskell, ... }:
+hfinal: hprev:
+let
+  inherit (haskell.lib.compose) markUnbroken;
+in
+{
+  hs-opentelemetry-propagator-datadog = markUnbroken hprev.hs-opentelemetry-propagator-datadog;
+  if-instance = markUnbroken hprev.if-instance;
+}
```
jkachmar, line 9: as of the latest nixpkgs unstable snapshot it looks like these need to be explicitly marked unbroken

.envrc
9999years: This should definitely not be in the `.envrc`: https://github.com/NixOS/nix/issues/9649

9999years: `--accept-flake-config` is unfortunately a huge footgun

jkachmar:
more than likely that I will restructure this away from `flake-parts` in the future, so as to avoid it being required downstream by anyone consuming the flake

PR #58, opened 2025-02-13
revert(haskell): downgrade `hs-opentelemetry-api`

jkachmar:
## description

downstream consumers have pinned `hs-opentelemetry-api` to 0.1.0.0, so we need to revert this version in our own package set (for CI) and go back to using the old `InstrumentationLibrary` API.


jkachmar:
hm, okay well I guess all the checks passed so... merged?

PR #59, opened 2025-02-18
drop 'flake-parts'

jkachmar:
## description

this keeps many of the changes from #56, but drops the dependency on `flake-parts` so that downstream consumers don't incur a dependency upon it (and, by extension, `nixpkgs` or `nixpkgs-lib`).

we can also avoid a dependency on `flake-utils` by defining a small utility module to define `forAllSystems`, which means that anyone depending on this project's overlays as a flake input can use it either as:

```nix
{
  inputs = {
    hs-temporal-sdk = {
      url = "github:MercuryTechnologies/hs-temporal-sdk";
      inputs.nixpkgs.follows = "unused";
      inputs.devenv.follows = "unused";
      inputs.fenix.follows = "unused";
    };
  };
}
```
...or, even more simply:
```nix
{
  inputs = {
    hs-temporal.sdk = {
      url = "github:MercuryTechnologies/hs-temporal-sdk";
      flake = false;
    };
  };
}
```


jkachmar:
> You have exhausted your monthly CI quota
RIP

nix/utils/flake.nix
```diff
@@ -0,0 +1,26 @@
...
+    nixpkgs.lib.genAttrs
+      [
+        "aarch64-darwin"
+        "aarch64-linux"
+        "x86_64-linux"
+      ]
+      (
+        system:
+        withPkgs {
+          pkgs = import nixpkgs {
```
jkachmar, line 15: by defining this as an attrset w/ `nixpkgs` as the imported `pkgs`, we can extend it (in the future) to include a matrix of Nix package sets that we want to test against.

e.g. `pkgs` -> nixpkgs-24.11, `darwinPkgs` -> `nixpkgs-darwin-24.11` `unstable` -> `nixpkgs-unstable`

flake.nix
```diff
@@ -12,14 +11,109 @@
...
+              modules = [
+                ./nix/devenv/temporal-bridge.nix
+                ./nix/devenv/temporal-dev-server.nix
+                (
+                  { pkgs, ... }:
+                  let
+                    haskellUtils = import ./nix/utils/haskell.nix pkgs;
+                  in
+                  {
+                    # XXX: Do we need to pull `proto-lens-protoc` from the
```
iand675, line 38: To answer this question, yeah, it needs to be the same. For example, we probably need to add support for JSON-encoding/decoding of protos to better support downloading and saving workflow histories from Temporal Cloud for regression tests and such. That will entail adding some new fields to the `proto-lens` type classes, some of which may not be optional. That will therefore necessitate that `proto-lens-protoc` knows about the new fields it needs to codegen.

flake.nix
```diff
@@ -12,14 +11,109 @@
...
+        haskellUtils.localPackageMatrix
+        // {
+          temporal-bridge = pkgs.temporal_bridge;
+        }
+      );
+
+      haskellOverlays = {
+        hs-temporal-sdk = import ./nix/overlays/haskell/hs-temporal-sdk.nix;
+        dependencies = {
+          default = import ./nix/overlays/haskell/deps.nix;
```
iand675, line 78: Maybe a small nit, but I'd like it if the default had an immediately obvious way to see what ghc version it is.

jkachmar: maybe this deserves a comment, but the idea here is that `haskellOverlays.dependencies.default` are the overrides that apply to _all_ GHC package sets.

`haskellOverlays.dependencies.ghc910` provides overrides that apply to GHC 9.10 on top of the defaults.

iand675: ah, got it.

PR #71, opened 2025-03-12
STAB-516: add support for api keys

johnmastro:
add support for using API keys.

this was smoke-tested by running mercurytechnologies/mercury-web-backend#42097 locally, pointing that at the (new, created for this purpose) `staging-ng` namespace, and successfully executing a workflow.

for the test, i started mwb with `TEMPORAL_ENABLED=true TEMPORAL_WORKER_ENABLED=true TEMPORAL_HOST="https://us-east-1.aws.api.temporal.io:7233" TEMPORAL_NAMESPACE="staging-ng.uoqel" TEMPORAL_API_KEY="<snip>" make run

Screenshot 2025-03-13 at 3 29 29 PM

### DIFF
--- a/core/rust/src/client.rs
+++ b/core/rust/src/client.rs
@@ -23,6 +23,7 @@ pub struct ClientConfig {
     identity: String,
     tls_config: Option<ClientTlsConfig>,
     retry_config: Option<ClientRetryConfig>,
+    api_key: Option<String>,
 }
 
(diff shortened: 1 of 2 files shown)


iand675 (commented):
I'd probably like to see a newtype for the APIKey type. Also, would be good if you could run something against a Temporal Cloud namespace to informally verify that it works.

iand675 (approved):

PR #75, opened 2025-03-26
add support for `update`

johnmastro:
Add support for the relatively new `update` operation ([announcement](https://temporal.io/blog/announcing-a-new-operation-workflow-update#introducing-workflow-update), [docs](https://docs.temporal.io/encyclopedia/workflow-message-passing))

We have some related backlog tickets and work-in-progress that we've decided shouldn't block shipping this initial pass:
- Support `startUpdate` vs `executeUpdate` semantics (STAB-885)
- Temporal OTEL interceptors propagate context too indiscriminately (STAB-846)
- Throw nicer errors from `Temporal.Core` where possible (STAB-859)
- Ian's work in #99 which will, among other things, enable some observability options for us


iand675 (approved):
Approved aside from the stray `print` call

sdk/src/Temporal/Common.hs
```diff
@@ -47,6 +47,11 @@ newtype WorkflowId = WorkflowId {rawWorkflowId :: Text}
   deriving newtype (Eq, Ord, Show, Hashable, IsString, ToJSON, FromJSON)
 
 
+newtype UpdateId = UpdateId {rawUpdateId :: Text}
+  deriving stock (Lift)
+  deriving newtype (Eq, Ord, Show, Hashable, IsString, ToJSON, FromJSON)
```
jkachmar, line 52: should `Eq, Ord, Show` be `stock`-derived? I feel like `Show` at least should be if we want it to be valid

sdk/src/Temporal/Contrib/OpenTelemetry.hs
```diff
@@ -227,6 +226,58 @@ makeOpenTelemetryInterceptor = do
...
+                -- context for it.
+                let spanArgs =
+                      defaultSpanArguments
+                        { kind = Server
+                        , attributes = mempty
+                        -- HashMap.fromList
+                        --   [ ("temporal.workflow_id", toAttribute $ rawWorkflowId $ input.handleQueryInputInfo.workflowId)
+                        --   , ("temporal.run_id", toAttribute $ rawRunId $ input.handleQueryInputInfo.runId)
+                        --   , ("temporal.workflow_type", toAttribute $ input.handleQueryInputType)
+                        --   ]
```
jkachmar, line 244: dead code?

sdk/src/Temporal/Contrib/OpenTelemetry.hs
```diff
@@ -227,6 +226,58 @@ makeOpenTelemetryInterceptor = do
...
+            , validateUpdate = \input next -> do
+                let spanArgs =
+                      defaultSpanArguments
+                        { kind = Server
+                        , attributes = mempty
+                        -- HashMap.fromList
+                        --   [ ("temporal.workflow_id", toAttribute $ rawWorkflowId $ input.handleQueryInputInfo.workflowId)
+                        --   , ("temporal.run_id", toAttribute $ rawRunId $ input.handleQueryInputInfo.runId)
+                        --   , ("temporal.workflow_type", toAttribute $ input.handleQueryInputType)
+                        --   ]
```
jkachmar, line 272: dead code?

sdk/src/Temporal/Common.hs
johnmastro: I paralleled the deriving strategies of `WorkflowId`, `SignalId`, etc, but I'm not sure what all considerations went into that - @iand675 can you weigh in here?

jkachmar:
```
ghci> newtype X = X { unX :: Text } deriving stock Show
ghci> X "hello"
ghci> newtype Y = Y { unY :: Text } deriving newtype Show
X {unX = "hello"}
ghci> Y "world"
"world"
```
that's really all it does.
---
for `Eq` & `Ord` I guess `newtype`-deriving means it's coerced so maybe there's a (very mild?) savings in terms of code size (idk?)

sdk/src/Temporal/Contrib/OpenTelemetry.hs
```diff
@@ -17,36 +17,34 @@ the interceptor to your Temporal client and worker configuration.
...
 
-
+import Control.Monad.Catch
 import Control.Monad.IO.Class
 import qualified Data.HashMap.Strict as HashMap
 import Data.Int
 import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
+-- TODO rework WorkflowExitVariant to not expose internals
+
```
jkachmar, line 27: I think this got moved & it should be around line 94?

sdk/src/Temporal/WorkflowInstance.hs
```diff
@@ -23,6 +23,8 @@ import qualified Control.Exception as E
 import Control.Monad
 import Control.Monad.Logger
 import Control.Monad.Reader
+
+
 #if __GLASGOW_HASKELL__ < 910
 import Data.Foldable (foldl')
 #endif
```
jkachmar, line 30: did the formatter add this?

sdk/src/Temporal/WorkflowInstance.hs
```diff
@@ -435,44 +576,51 @@ applyJobs jobs fAwait = UnliftIO.try $ do
...
     )
       <$> fAwait
   where
     jobGroups =
       V.foldr
-        ( \job tup@(!patchNotifications, !signalWorkflows, !queryWorkflows, !resolutions, !otherJobs) -> case job ^. Activation.maybe'variant of
-            Just (WorkflowActivationJob'NotifyHasPatch n) -> (applyNotifyHasPatch n *> patchNotifications, signalWorkflows, queryWorkflows, resolutions, otherJobs)
-            Just (WorkflowActivationJob'SignalWorkflow sig) -> (patchNotifications, applySignalWorkflow sig : signalWorkflows, queryWorkflows, resolutions, otherJobs)
-            Just (WorkflowActivationJob'QueryWorkflow q) -> (patchNotifications, signalWorkflows, applyQueryWorkflow q *> queryWorkflows, resolutions, otherJobs)
+        ( \job jobGroups -> case job ^. Activation.maybe'variant of
```
jkachmar, line 592: we lost the strictness annotations on this tuple unpacking, but `JobGroups` above doesn't have annotations on its fields.

sdk/test/IntegrationSpec.hs
```diff
@@ -1538,6 +1552,493 @@ needsClient = do
       incompatibleReplayResult <- runReplayHistory globalRuntime incompatibleConf history
       incompatibleReplayResult `shouldSatisfy` isLeft
 
+  describe "Update" $ do
+    it "works with no validator" $ \TestEnv {..} -> do
+      let conf = provideCallStack $ configure () (discoverDefinitions @() $$(discoverInstances) $$(discoverInstances)) $ do
```
jkachmar, line 1557: are there any drawbacks to using TH instance discovery here?

`discoverInstances` is going to look up all `ActivityFn` & `WorkflowFn` in scope that match `env ~ ()`; is there a reason not to do this (config segmentation between test cases) or is it all good?

sdk/src/Temporal/Common.hs
johnmastro: what's the validity concern for `Show`?

sdk/src/Temporal/WorkflowInstance.hs
johnmastro: yeah, this is a little weird. if i try to commit with no empty lines there, fourmolu prints this diff adding a blank line, followed by an `AST of input and AST of formatted code differ` error:

```
sdk/src/Temporal/WorkflowInstance.hs
@@ -23,6 +23,7 @@
  import Control.Monad
  import Control.Monad.Logger
  import Control.Monad.Reader
+
  #if __GLASGOW_HASKELL__ < 910
  import Data.Foldable (foldl')
  #endif

  AST of input and AST of formatted code differ.
  Please, consider reporting the bug.
  To format anyway, use --unsafe.
```

if i add a single blank line (as shown in the diff), then on the next run fourmolu adds the second blank line (but doesn't report an error).

sdk/src/Temporal/Contrib/OpenTelemetry.hs
johnmastro: indeed, just pushed a commit moving it back

sdk/src/Temporal/WorkflowInstance.hs
johnmastro: just pushed a commit adding strictness annotations on `JobGroup`'s fields

sdk/test/IntegrationSpec.hs
iand675: There's not really any downside. Obviously, we could do a bit of factoring to call this in one spot to speed things up by not reevaluating the TH, but I feel like it's probably better to tackle that outside of this PR

PR #85, opened 2025-04-11
Add unsafe functions for o11y needs

iand675:
We want to be able to support logging, tracing, metrics, error reporting, etc. within Workflows. At the moment we can pretty much only achieve this via interceptors, which means that we can't, for example, add attributes or create spans within a workflow.

So, we're adding an escape hatch here that needs to be used relatively judiciously, but gives us a starting point.

### DIFF
--- a/sdk/src/Temporal/Workflow/Unsafe.hs
+++ b/sdk/src/Temporal/Workflow/Unsafe.hs
@@ -1,0 +1,59 @@
+{- | Unsafe functions for performing IO in the Workflow monad.
+
+These functions fundamentally break the purity of the Workflow monad, so
+they should be used with extreme caution.
+
+In particular, they should not be used to alter the control flow of the
+Workflow.
+
+The reason that they exist at all is because there are some operations
+needed for observability and debugging that don't produce observable changes
+to the Workflow state, such as logging.
+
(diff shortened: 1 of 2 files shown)


sdk/src/Temporal/Workflow/Unsafe.hs
```diff
@@ -0,0 +1,59 @@
...
+Temporal SDK:
+
+* For time, use 'Workflow.currentTimeMillis'
+* For randomness, use 'Workflow.Random'
+* For external data, use Activities
+* For side effects, use 'Workflow.sideEffect'
+-}
+performUnsafeNonDeterministicIO :: IO a -> Workflow a
+performUnsafeNonDeterministicIO m = Workflow (\_ -> Done <$> liftIO m)
+{-# INLINE performUnsafeNonDeterministicIO #-}
```
jkachmar, line 59: it's silly but years of carefully making sure that `unsafePerformIO[..]` functions were `NOINLINE`-d made me stop and read this line like 3 times

jkachmar (approved):
good to at least have this escape hatch; I suppose we can start putting otel helpers in a contrib module once we figure out an interface that works?

iand675:
That's the plan!

PR #86, opened 2025-04-11
Add some new time utilities

iand675:
We should have some sleepUntil type functions instead of making people do the math themselves with UTCTimes.

### DIFF
--- a/sdk/src/Temporal/Duration.hs
+++ b/sdk/src/Temporal/Duration.hs
@@ -39,6 +39,8 @@ module Temporal.Duration (
   durationFromProto,
   durationToProto,
   durationToMilliseconds,
+  addDurationToSystemTime,
+  diffSystemTime,
 ) where
 
 import Data.Aeson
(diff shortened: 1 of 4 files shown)


sdk/src/Temporal/Duration.hs
```diff
@@ -217,3 +220,36 @@ human-readable values emitted by logs, metrics, traces, etc.
...
+addDurationToSystemTime (Duration durSecs durNanos) (MkSystemTime sysSecs sysNanos) =
+  let
+    -- Convert Word32 to Int64 for safe addition
+    totalNanos = fromIntegral sysNanos + fromIntegral durNanos
+    -- Normalize the result to handle nanosecond overflow/underflow
+    (# overflowSecs, finalNanos #) = normalize 0 totalNanos
+    -- Add all seconds together
+    finalSecs = sysSecs + durSecs + overflowSecs
+  in
+    MkSystemTime finalSecs (fromIntegral finalNanos)
```
jkachmar, line 239: man we really need this functionality in a general purpose library (if not in `time` itself), huh?

jkachmar (approved):
🎉

PR #88, opened 2025-04-11
STAB-818: add haskell-language-server in devenv

johnmastro:


nix/devenv/haskell.nix
```diff
@@ -1,16 +1,13 @@
-ghcVersion:
```
johnmastro, line 1: this is from the precommit formatter 🤷‍♂️

matto-mercury (approved):

PR #91, opened 2025-04-18
define default-extensions centrally

johnmastro:
I think it would be nice, for DX-y reasons, to define things like `default-extensions` and `ghc-options` centrally for this repo, rather than separately in each package's `package.yaml`.

This PR starts that process by adopting [mwb's strategy](https://github.com/MercuryTechnologies/mercury-web-backend/tree/master/hpack-common) of [including](https://github.com/MercuryTechnologies/mercury-web-backend/blob/4024c75d8c68c31c6f02c9a0fdc0ef960e723378/package.yaml#L1) an `hpack-common/default.yaml` into `package.yaml`s. To start, it only takes over `default-extensions`, though my intention is to add at least `ghc-options` as well.

Some examples of why I think this would be useful:
- We have a number of warnings like `missing-role-annotations` and `missing-poly-kind-signatures` in here that I believe are essentially spurious (in the sense that we don't intend to act on them). It would be nice to disable them, and in that case it would be nice to do that in one place (as [mwb does](https://github.com/MercuryTechnologies/mercury-web-backend/blob/master/hpack-common/ghc-options.yaml#L66) for those same warnings)
- I'm currently using hls in this repo mostly-successfully, but I'd love to get the `ghciwatch`/`static-ls` workflow set up in here eventually, which requires some particular compiler options. As above, it would be nice to configure that in one place
- `fourmolu` needs to know what extensions are active (to handle e.g. record dots correctly), so it's convenient to have that defined in a single place to make things like format-on-save work correctly. For example, in mwb's `run-format.sh`, [we rely on default extensions being defined in a single well-known location](https://github.com/MercuryTechnologies/mercury-web-backend/blob/4024c75d8c68c31c6f02c9a0fdc0ef960e723378/nix/packages/mercury/run-format.sh#L65-L67) (relative to repo root) so we can pass them on.

### DIFF
--- a/codec-encryption/package.yaml
+++ b/codec-encryption/package.yaml
@@ -1,3 +1,5 @@
+<<: !include hpack-common/defaults.yaml
+
 verbatim: { cabal-version: 3.0 }
 
 name:         temporal-codec-encryption
@@ -19,27 +21,6 @@ dependencies:
 - proto-lens
 - crypton
 - memory
-
-default-extensions:
-- DataKinds
-- FlexibleContexts
-- FlexibleInstances
-- GADTs
-- GeneralizedNewtypeDeriving
-- InstanceSigs
-- LambdaCase
-- MultiParamTypeClasses
-- NamedFieldPuns
-- RankNTypes
-- OverloadedRecordDot
-- OverloadedStrings
-- RecordWildCards
-- ScopedTypeVariables
-- StandaloneDeriving
-- TypeApplications
-- TypeFamilies
-- UndecidableInstances
-- TypeOperators
 
 ghc-options:
 # For details on warnings: https://downloads.haskell.org/~ghc/master/users-guide/using-warnings.html
--- a/codec-encryption/package.yaml
+++ b/codec-encryption/package.yaml
@@ -1,4 +1,5 @@
-<<: !include hpack-common/defaults.yaml
+defaults:
+  local: ../hpack-common/defaults.yaml
 
 verbatim: { cabal-version: 3.0 }
 
--- a/hpack-common/defaults.yaml
+++ b/hpack-common/defaults.yaml
@@ -1,1 +1,2 @@
-<<: !include default-extensions.yaml
+_default-extensions: !include "default-extensions.yaml"
+default-extensions: *common-default-extensions
--- a/hpack-common/defaults.yaml
+++ b/hpack-common/defaults.yaml
@@ -1,0 +1,1 @@
+<<: !include default-extensions.yaml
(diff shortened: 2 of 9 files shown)


codec-encryption/package.yaml
```diff
@@ -1,3 +1,5 @@
+<<: !include hpack-common/defaults.yaml
+
```
jkachmar, line 2: `hpack` supports including [common fields](https://github.com/sol/hpack?tab=readme-ov-file#common-fields) using a [defaults] stanza.

e.g.
```yaml
defaults:
  local: ../hpack/defaults.yaml
```
...and then you could define `defaults.yaml` as something like:
```yaml
_default-extensions: !include ".default-extensions.yaml"
default-extensions: *common-default-extensions
```
...and `default-extensions.yaml` as:
```yaml
default-extensions: &common-default-extensions
  - NoImplicitPrelude
  - OverloadedStrings
  - ScopedTypeVariables
```
...if you wanted to keep everything broken out like this.

I have a (very old) example of this sort of organization in case you decide to go down this route:
- [defaults](https://github.com/jkachmar/dry-it-up/blob/master/package-defaults.yaml)
- [extensions anchor](https://github.com/jkachmar/dry-it-up/blob/master/.stylish-haskell.yaml)
- [use-site](https://github.com/jkachmar/dry-it-up/blob/master/foo/package.yaml)

jkachmar:
this is good stuff; we should also migrate [core/temporal-sdk-core.cabal](https://github.com/MercuryTechnologies/hs-temporal-sdk/tree/main/core/temporal-sdk-core.cabal) from `.cabal` to `package.yaml` but that can be in a different PR

jkachmar (approved):

hpack-common/defaults.yaml
```diff
@@ -0,0 +1,2 @@
+_default-extensions: !include "default-extensions.yaml"
+default-extensions: *common-default-extensions
```
jkachmar, line 2: fwiw, it might also make sense to just list the `default-extensions` directly in this file; the only reason it's worth factoring out into a separate YAML file is if we want to reuse the `default-extensions.yaml` file for another purpose.

optimal-codec/package.yaml
```diff
@@ -1,3 +1,6 @@
+defaults:
+  local: ../hpack-common/defaults.yaml
+
 verbatim: { cabal-version: 3.0 }
```
jkachmar, line 4: I think `verbatim` also works in the common file, which we also want to repeat across all these `package.yaml`s as well

PR #93, opened 2025-04-22
hpack-ify temporal-sdk-core

jkachmar:
## description

builds off of #91 (but kind of steps on the toes of #92, so maybe some of this gets changed a bit)


core/temporal-sdk-core.cabal
```diff
@@ -1,86 +1,107 @@
...
+      OverloadedRecordDot
+      OverloadedStrings
+      RecordWildCards
+      ScopedTypeVariables
+      StandaloneDeriving
+      TypeApplications
+      TypeFamilies
+      TypeOperators
+      UndecidableInstances
+      ViewPatterns
```
jkachmar, line 78: note: check that this is actually okay; if not then it probably makes sense to swap back to using YAML anchors/imports

PR #94, opened 2025-04-22
re-run protobuf codegen

jkachmar:
## description

we've drifted a bit relative to upstream; this updates protobuf code to track the latest commit from `sdk-core` & fixes a breaking change around workflow activation (renamed from "start" -> "initialize").


sdk/src/Temporal/WorkflowInstance.hs
```diff
@@ -464,10 +463,12 @@ applyJobs jobs fAwait = UnliftIO.try $ do
             Just (WorkflowActivationJob'UpdateRandomSeed updateRandomSeed) -> (patchNotifications, signalWorkflows, queryWorkflows, resolutions, otherJobs *> applyUpdateRandomSeed updateRandomSeed)
             Just (WorkflowActivationJob'CancelWorkflow cancelWorkflow) -> (patchNotifications, signalWorkflows, queryWorkflows, PendingWorkflowCancellation cancelWorkflow : resolutions, otherJobs)
             -- By the time we get here, the workflow should already be running.
-            Just (WorkflowActivationJob'StartWorkflow _startWorkflow) -> tup
+            Just (WorkflowActivationJob'InitializeWorkflow _initializeWorkflow) -> tup
             -- Handled in the worker.
             Just (WorkflowActivationJob'RemoveFromCache _removeFromCache) -> tup
             Just (WorkflowActivationJob'DoUpdate _) -> error "DoUpdate not yet implemented"
+            Just (WorkflowActivationJob'ResolveNexusOperation _) -> error "ResolveNexusOperation not yet implemented"
+            Just (WorkflowActivationJob'ResolveNexusOperationStart _) -> error "ResolveNexusOperationStart not yet implemented"
```
jkachmar, line 474: I think this should be fine? maybe worth a comment that this should be impossible to encounter given that the library doesn't implement any function which would produce this activation

sdk/src/Temporal/WorkflowInstance.hs
```diff
@@ -77,7 +78,9 @@ import Temporal.Workflow.Internal.Monad
 import Temporal.Workflow.Types
 import UnliftIO
 import Data.Vault.Strict (Vault)
-
+#if __GLASGOW_HASKELL__ < 910
+import Data.Foldable (foldl')
+#endif
```
jkachmar, line 83: this (& `LANGUAGE CPP`) goes away once we drop support for older GHC versions

iand675 (approved):

PR #95, opened 2025-04-23
update `temporal_bridge` deps

jkachmar:
## description

based off of #94; this bumps deps in `temporal_bridge` & marks (most of) them as tolerant of patch bumps w/ `~`


jkachmar:
next step after this is going to be bumping `sdk-core`, which will require some changes on the Rust side to account for [WorkerVersioningStrategy](https://github.com/temporalio/sdk-core/blob/6f2dd3743e788f81afa839d7a423007ce4e6cb46/core-api/src/worker.rs#L522), [WorkerConfig.workflow_task_poller_behavior](https://github.com/temporalio/sdk-core/blob/6f2dd3743e788f81afa839d7a423007ce4e6cb46/core-api/src/worker.rs#L41-L45), and a few others.

core/rust/Cargo.toml
```diff
@@ -14,20 +14,17 @@ temporal-sdk-core = { git = "https://github.com/temporalio/sdk-core", rev = "6e6
...
+tokio-stream = "~0.1"
+tonic = "~0.12"
+tracing = "~0.1"
+url = "~2.5"
+ffi-convert = "~0.6"
+serde = { version = "~1.0", features = ["derive"] }
+serde_json = "~1.0"
+prost = "~0.13"
+prost-types = "~0.13"
+libc = "0.2.172"
```
jkachmar, line 27: @iand675 sanity check: it feels like it's safe for us to put `~` range bounds on these yeah?

iand675: Yeah, seems fine to me

iand675 (approved):

PR #98, opened 2025-04-29
STAB-506: add first execution run ID to WorkflowHandle

johnmastro:
this is a pre-factor for STAB-506 (adding `update`) that i pulled out for easier review.

in the course of STAB-506, we noticed we were handling `firstExecutionRunId` (which we're currently taking as a parameter to `terminate` - the other main operation that, like `update`, supports it) differently than the typescript sdk (which takes it as an argument to `getHandle`, and then makes it a property on the workflow handle, such that later operations on that handle use it automatically[^1]).

so, adopt the typescript sdk's approach. in the main STAB-506 pr we can then build on this and have `update` function in the same way.

[^1]: well, some operations (like `signal` and `query`) just totally ignore it. i asked about this on the temporal slack and they said it's a known gap that was never prioritized.

### DIFF
--- a/sdk/src/Temporal/Client.hs
+++ b/sdk/src/Temporal/Client.hs
@@ -33,7 +33,7 @@ module Temporal.Client (
   startWorkflowOptions,
   Temporal.Client.start,
   startFromPayloads,
-  WorkflowHandle,
+  WorkflowHandle (..),
   execute,
   waitWorkflowResult,
 
@@ -58,6 +58,7 @@ module Temporal.Client (
 
   -- * Producing handles for existing workflows
   getHandle,
+  GetHandleOptions (..),
 
   -- * Workflow history utilities
   fetchHistory,
@@ -255,7 +256,7 @@ This function will block until the workflow completes, and will return the resul
 or throw a 'WorkflowExecutionClosed' exception if the workflow was closed without returning a result.
 -}
 waitWorkflowResult :: (Typeable a, MonadIO m) => WorkflowHandle a -> m a
-waitWorkflowResult h@(WorkflowHandle readResult _ c wf r) = do
+waitWorkflowResult h@(WorkflowHandle readResult _ c wf r _) = do
   mev <- runReaderT (waitResult wf r c.clientConfig.namespace) c
   case mev of
     Nothing -> error "Unexpected empty history"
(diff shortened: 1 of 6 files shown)


sdk/src/Temporal/Client.hs
```diff
@@ -459,8 +469,9 @@ getHandle
   => KnownWorkflow args a
   -> WorkflowId
   -> Maybe RunId
+  -> Maybe GetHandleOptions
   -> m (WorkflowHandle a)
-getHandle (KnownWorkflow {knownWorkflowCodec, knownWorkflowName}) wfId runId = do
+getHandle (KnownWorkflow {knownWorkflowCodec, knownWorkflowName}) wfId runId opts = do
```
johnmastro, line 474: i'm mirroring [how the typescript sdk handles these arguments](https://github.com/temporalio/sdk-typescript/blob/5ff22f0f7f0ac9ed543bbd5dea3f6205cadafff5/packages/client/src/workflow-client.ts#L1516-L1517) (`runId` is an optional top-level parameter, while `firstExecutionRunId` is embedded in an options object) but i'm not sure i like it.

alternatively, we could move `runId` into `GetHandleOptions` (alongside `firstExecutionRunId`).

i'd be interested to hear what others think.

sdk/test/IntegrationSpec.hs
```diff
@@ -965,6 +967,163 @@ needsClient = do
           useClient (C.execute parentWf.reference (W.WorkflowId parentId) opts)
             `shouldReturn` "Left ChildWorkflowCancelled"
 
+    describe "Terminate" $ do
```
johnmastro, line 970: this is perhaps a lot of tests just for this one aspect of `terminate`, but i wanted to convince myself i understood the semantics of `runId` and `firstExecutionRunId` and especially combining them

iand675 (approved):
Approved, but it might be nice if the integration tests could be in their own function definition and pulled into the larger test block

sdk/test/IntegrationSpec/NoOpWorkflow.hs
```diff
@@ -0,0 +1,25 @@
...
+import Control.Monad
+import RequireCallStack (provideCallStack)
+import Temporal.Duration
+import Temporal.Payload
+import Temporal.TH
+import Temporal.Workflow
+
+
+noOpWorkflow :: Workflow ()
+noOpWorkflow = provideCallStack $ pure ()
```
jkachmar, line 22: I think we can inline this in the test case (or at least test module) itself yeah?
```haskell
let nop :: RequireCallStack => Workflow ()
    nop = pure ()
    nopWf = provideWorkflow defaultCodec "nop-workflow" nop
    nopRef = nopWf.reference
    nopDef = nopWf.definition
in ...

johnmastro: yeah definitely. that's what i did first, but in the end breaking it out and importing it felt a little cleaner to me. i don't have a strong preference though

PR #99, opened 2025-04-29
Overhaul workflow monad to use native threading

iand675:
This is a relatively experimental change, but it's intended to enhance our ability to add new code in the future. The `Coroutine` monad, and the `Workflow` monad implementation are currently fairly hard to add new instances for, hard to write interceptors for, and lead to a lot of questions from engineers about the semantics of threads and exceptions in the Temporal model (as it currently stands). This PR is a step in the direction of using Haskell-native threads + thin wrappers around the IO monad to run Workflow code. This should provide us with the ability to write instances that we more-or-less _can't_ write at the moment. The crucial one is that we can't reliably handle spans from OTel due to not having a good `bracket`-like function in the face of continuations.


.envrc
iand675, line 1: pre-commit check fix.

.envrc
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
```
iand675, line 1: Precommit check fix

sdk/src/Temporal/Interceptor.hs
```diff
@@ -144,3 +152,107 @@ instance Monoid ScheduleClientInterceptors where
     ScheduleClientInterceptors
       { scheduleWorkflowAction = \opts _ -> return opts
       }
+
+
+data ExecuteWorkflowInput = ExecuteWorkflowInput
```
iand675, line 157: Moved this stuff here from elsewhere to make the `Temporal.Interceptor` module the main interface for what goes in and out of Temporal actions

sdk/package.yaml
```diff
@@ -87,17 +87,17 @@ library:
...
   - Temporal.WorkflowInstance
   - Temporal.Workflow.Worker
-  - Temporal.Workflow.Eval
   - Temporal.Workflow.WorkflowInstance
   - Temporal.Worker.Types
   - Temporal.Activity.Types
+  - Temporal.Workflow.ChildWorkflowHandle
+  - Temporal.Workflow.CommandQueue
   - Temporal.Workflow.Unsafe.Handle
-  - Temporal.Workflow.Internal.Instance
```
iand675, line 99: Moving away from the internal namespace usage since we can just not expose the modules

sdk/src/Temporal/Workflow.hs
```diff
@@ -200,9 +198,6 @@ module Temporal.Workflow (
   ProvidedUpdate (..),
   KnownUpdate (..),
 
-  -- * Other utilities
-  unsafeAsyncEffectSink,
```
iand675, line 204: Not used, not really needed any more.

sdk/src/Temporal/Workflow.hs
```diff
@@ -1644,12 +1472,12 @@ returning the original data structure with the arguments replaced with the resul
 This is actually a bit of a misnomer, since it's really 'traverseConcurrently', but this is copied to mimic the 'async' package's naming
 to slightly ease adoption.
 -}
-mapConcurrently :: Traversable t => (a -> Workflow b) -> t a -> Workflow (t b)
+mapConcurrently :: (Traversable t) => (a -> Workflow b) -> t a -> Workflow (t b)
 mapConcurrently = traverseConcurrently
```
iand675, line 1476: I have a feeling that these folding / traversing versions of this would benefit from a specialized version that does a lot less blocking, but probably just need to come back to that later if it's an actual problem

sdk/src/Temporal/Workflow/Instance.hs
```diff
@@ -0,0 +1,225 @@
...
+
+
+data WorkflowRuntime = WorkflowRuntime
+  { workflowRuntimeInstance :: {-# UNPACK #-} !WorkflowInstance
+  , workflowRuntimeThreads :: {-# UNPACK #-} !ThreadManager
+  , workflowRuntimeCommandQueue :: {-# UNPACK #-} !CommandQueue
+  , workflowRuntimeCancelRequested :: {-# UNPACK #-} !(IVar ())
+  , workflowRuntimeSequenceMaps :: {-# UNPACK #-} !SequenceMaps
+  , workflowRuntimeUnappliedJobs :: {-# UNPACK #-} !(TVar Int)
+  , signals :: {-# UNPACK #-} !SignalSupport
```
iand675, line 81: This is an approach I'm taking on trying to decompose portions of workflow functionality into smaller units that are a bit more independently maintainable. Hopefully this unlocks some ability to test subsystems better in future PRs.

sdk/test/IntegrationSpec/HangingWorkflow.hs
```diff
@@ -81,22 +81,6 @@ defaultActivityOptions :: StartActivityOptions
 defaultActivityOptions = defaultStartActivityOptions $ StartToClose $ minutes 1
 
 
-opts :: StartChildWorkflowOptions
-opts =
```
iand675, line 85: Moved to not be a global value to fix shadowing warnings

sdk/test/IntegrationSpec.hs
```diff
@@ -2191,13 +2178,11 @@ terminateTests = do
...
                       (workflowRef NoOpWorkflow)
                       "test-terminate-throws-with-good-run-id-bad-first-execution-run-id"
-                      (Just $ fromJust h.workflowHandleRunId)
-                      (Just C.GetHandleOptions {C.firstExecutionRunId = Just "bad-first-execution-run-id"})
+                      (C.GetHandleOptions {C.runId = Just $ fromJust h.workflowHandleRunId, C.firstExecutionRunId = Just "bad-first-execution-run-id"})
                   C.terminate h' C.TerminationOptions {terminationReason = "testing", terminationDetails = []}
               )
             `shouldThrow` \case
               RpcError {} -> True
-              _ -> False
```
iand675, line 2200: Unnecessary since RPCError only has one constructor

sdk/src/Temporal/Workflow/Worker.hs
```diff
@@ -123,22 +118,14 @@ execute worker@WorkflowWorker {workerCore} = flip runReaderT worker $ do
...
-          $(logDebug) $ Text.pack ("Got activation " <> show activation)
-          -- We want to handle activations as fast as possible, so we don't want to block
-          -- on dispatching jobs. We link the activator thread to the run-loop so that any
-          -- unhandled exceptions in that logic aren't ignored.
-          activationCtxt <- getContext
-          activator <- asyncLabelled (Text.unpack $ Text.concat ["temporal/worker/workflow/activate", Core.namespace c, "/", Core.taskQueue c]) $ do
-            _ <- attachContext activationCtxt
-            handleActivation activation
-          link activator
+          handleActivation activation
```
iand675, line 128: We don't need this extra bit of async calling, because creating the instances is cheap, and the `RunningWorkflow` struct is in charge of its own async-ness now.

sdk/src/Temporal/Client.hs
```diff
@@ -275,15 +272,9 @@ waitWorkflowResult h@(WorkflowHandle readResult _ c wf r _) = do
...
           let payloads = convertFromProtoPayload <$> (attrs ^. History.result . Common.payloads)
-          -- LMAO this is such a comical coercion
-          --
-          -- We just need to ignore payloads if the result type is unit to allow workflows to evolve
-          -- such that they can return something else in the future.
-          if typeRep h == typeOf ()
-            then pure $ unsafeCoerce ()
-            else case payloads of
-              (a : _) -> liftIO $ readResult a
-              _ -> error "Missing result payload"
```
jkachmar, line 286: nice

sdk/src/Temporal/Client.hs
```diff
@@ -574,23 +563,15 @@ startFromPayloads k@(KnownWorkflow codec _) wfId opts payloads = do
...
-      }
+        runReaderT
+          ( do
+              let r = RunId $ swer ^. WF.runId
+              getHandle k wfId' $
+                GetHandleOptions
+                  (Just r)
+                  (Just r)
+          )
+          c
```
jkachmar, line 574: the world if Haskell had a `runReaderT` variant whose arguments were flipped by default

sdk/src/Temporal/Client.hs
```diff
@@ -521,30 +512,28 @@ startFromPayloads
   -> StartWorkflowOptions
   -> V.Vector UnencodedPayload
   -> m (WorkflowHandle result)
-startFromPayloads k@(KnownWorkflow codec _) wfId opts payloads = do
+startFromPayloads k wfId opts payloads = do
```
jkachmar, line 515: not something to change in code, but would it be useful to have a glossary for stuff like "`k` is commonly used to refer to the continuation function" type stuff (for newcomers)?

sdk/src/Temporal/Workflow.hs
```diff
@@ -540,43 +514,46 @@ signalWorkflow
...
+        defMessage
+          & Command.signalExternalWorkflowExecution
+            .~ f
+              ( defMessage
+                  & Command.seq .~ rawSequence s
+                  & Command.signalName .~ signalName
+                  & Command.vec'args .~ fmap convertToProtoPayload args
+                  -- TODO
+                  -- & Command.headers .~ _
+              )
```
jkachmar, line 535: I know this is just a carryover, but what's the TODO here for? just setting the `Command.headers` to something if/when we want to support that?

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+import qualified Data.Map.Strict as Map
+import Data.Maybe
+import Data.ProtoLens
+import qualified Data.Set as Set
+import qualified Data.Text as Text
+import Data.Traversable
+import Data.Vector (Vector)
+import Data.Word
+import GHC.Conc (unsafeIOToSTM)
+import GHC.Exts (Any)
```
jkachmar, line 31: is this import actually used? I can't see `Any` referenced in the module at all

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+  readStateVar = readTVarIO . stateVarRef
+
+
+instance MonadReadStateVar Query where
+  readStateVar = Query . readTVar . stateVarRef
+
+
+instance MonadReadStateVar Condition where
+  readStateVar ref = Condition do
+    tid <- unsafeIOToSTM myThreadId
```
jkachmar, line 261: might be worth documenting this the same way that Rust usage of `unsafe` has a convention of documenting the invariants (e.g. `IO` and `STM` are effectively identical with respect to RTS thread ID)

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+
+      High level overview:
+
+      The main thread runs the core workflow implementation.
+      It is responsible for launching other worker threads, adding commands to the command queue.
+
+      There is a scheduler thread that waits for all threads to block, then signals that we should flush.
+      When we get an activation, we apply the jobs, then
+
+      -}
```
jkachmar, line 479: was there more here, or was this meant to be covered in other docs?

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+          deregisterThread runtime myTid
+          readTVar $ threadManager $ getThreadManager runtime
+        -- TODO, report abandoned signals and updates to user
+        mapM_ cancelWorkflowThreadNoWait $ HashMap.keys ts
+
+      eval :: InstanceM (WorkflowExitVariant Payload)
+      eval = mask $ \restore -> do
+        tid <- myThreadId
+        let tidText = Text.pack (show tid)
+            threadWrapper _restore = id
```
jkachmar, line 502: is the intent here the same as your previous `threadWrapper`? we could use this as a point to support propagating trace context?

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+
+      applyActivations mainThreadId = do
+        activation <- readTQueue runtime.workflowRuntimeInstance.activationChannel
+        activate runtime activation mainThreadId
+
+      applyQueuedSignals :: Async (WorkflowExitVariant Payload) -> STM (Int, IO ())
+      applyQueuedSignals mainThread = do
+        signalHandlers <- readTVar runtime.signals.signalHandlers
+        handleableSignals <- stateTVar runtime.signals.bufferedSignals $ \signals ->
+          span (\signal -> (Just (signal ^. Activation.signalName) `HashMap.member` signalHandlers) || (Nothing `HashMap.member` signalHandlers)) signals
```
jkachmar, line 520: TIL `stateTVar` + `span` as a pattern for handling state transition

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+        queryHandlers <- readTVar runtime.queries.queryHandlers
+        queriesToProcess <- stateTVar runtime.queries.bufferedQueries $ \queries -> (queries, [])
+        let (goodQueries, badQueries) =
+              span
+                ( \query ->
+                    (Just (query ^. Activation.queryType) `HashMap.member` queryHandlers) || (Nothing `HashMap.member` queryHandlers)
+                )
+                queriesToProcess
+        -- TODO, do we need to count unapplied jobs here. I think so.
+        modifyTVar' runtime.workflowRuntimeUnappliedJobs (+ length goodQueries)
```
jkachmar, line 540: not 100% sure what this comment is referring to: do you want to count unapplied jobs in one of the query counter variables here?

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+                  waitAllThreadsBlocked runtime
+                pure (handleCount == 0)
+              -- By the time we've applied all the signals that we can,
+              -- all of the queries we can possibly handle must have been
+              -- defined.
+              liftIO . snd =<< atomically (applyQueuedQueries mainThread)
+              restore $ atomically do
+                waitAllJobsHandled runtime
+                waitAllThreadsBlocked runtime
+            )
```
jkachmar, line 613: I think I've mentioned this elsewhere, but it may be worth reiterating to the reader (either here or elsewhere in the documentation) that `atomically` is being called explicitly to ensure a deliberate transactional sequence.

I think a lot of people can end up with an impression that `atomically` is just the `STM a -> IO a` function — and most of the time that's no big deal — but in this case if someone were to merge some of these blocks then you could end up with re-ordering that breaks the semantics we want to enforce.

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+      completionMessage =
+        defMessage
+          & Completion.runId .~ rawRunId info.runId
+          & Completion.successful .~ completionSuccessful
+  res <- liftIO $ runtime.workflowRuntimeInstance.workflowCompleteActivation completionMessage
+  case res of
+    Left err -> do
+      throwIO err
+    Right () -> do
+      pure ()
```
jkachmar, line 649: `either throwM pure res`? or is it worth deliberately `case`-ing & using `throwIO`

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+
+handleSignalWorkflow :: SignalWorkflow -> ApplyM ()
+handleSignalWorkflow sig = do
+  runtime <- askRuntime
+  let inst = runtime.workflowRuntimeInstance
+  handlers <- applySTM $ readTVar runtime.signals.signalHandlers
+  case specificHandlerOrDefault handlers (sig ^. Activation.signalName) of
+    Nothing -> applySTM do
+      -- TODO, if there's a huge amount of signals, this is algorithmically inefficient.
+      modifyTVar' runtime.signals.bufferedSignals (++ [sig])
```
jkachmar, line 776: what would the alternative be here, `Vector`?

iand675: The alternative would be to use a difference list, such as this, although they're not hard to implement by hand: https://hackage.haskell.org/package/dlist

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+      completeUpdate payload =
+        CommandQueue.addCommand runtime $
+          defMessage
+            & Command.updateResponse
+              .~ ( defMessage
+                    & Command.protocolInstanceId .~ (doUpdate ^. Activation.protocolInstanceId)
+                    & Command.completed .~ convertToProtoPayload payload
+                 )
+
+  -- (validator, updateAction) <- do
```
jkachmar, line 994: dead code?

sdk/test/Temporal/Workflow/IVarSpec.hs
```diff
@@ -0,0 +1,201 @@
...
+newTestRuntime :: IO TestRuntime
+newTestRuntime = do
+  threadManagerVar <- newTVarIO HashMap.empty
+  pure
+    TestRuntime
+      { testThreadManager = threadManagerVar
+      }
+
+
+-- -- | Creates a new thread in the test runtime
```
jkachmar, line 48:
```suggestion
-- | Creates a new thread in the test runtime
```

sdk/test/Temporal/Workflow/IVarSpec.hs
```diff
@@ -0,0 +1,201 @@
...
+    tid <- myThreadId
+    atomically $ do
+      workflowThreadBlocked <- newTVar False
+      modifyTVar' runtime.testThreadManager $
+        HashMap.insert tid (WorkflowThread workflowThreadBlocked)
+    runTestMonad runtime action
+      `finally` atomically (modifyTVar' runtime.testThreadManager $ HashMap.delete tid)
+
+
+-- -- | Check if a thread is blocked in the runtime
```
jkachmar, line 61:
```suggestion
-- | Check if a thread is blocked in the runtime
```

sdk/src/Temporal/Workflow/Monad.hs
```diff
@@ -0,0 +1,1163 @@
...
+addCommand command = do
+  inst <- ask
+  atomically $ CommandQueue.addCommand inst command
+
+
+-- Bit of a hack. This needs to be called for each workflow activity in the official SDK
+updateCallStack :: RequireCallStack => InstanceM ()
+updateCallStack = do
+  inst <- asks workflowRuntimeInstance
+  writeIORef inst.workflowCallStack $ popCallStack callStack
```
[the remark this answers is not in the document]
iand675: That's correct

sdk/src/Temporal/Workflow/Monad.hs
jkachmar: ah, so you mean the specific case of like... a lot of signals getting queued up before any of them get handled yea?

iand675: Right, it would be an N^2 evaluation of the list. Hard to think that gRPC size limits would let this get too out of control though

jkachmar (approved):
huge quality of life improvement; I wish we had some benchmarks to do a before/after comparison because I'd be pretty shocked if this didn't measurably improve overhead.

iand675:
@jkachmar @andrewsmith Informally benchmarking this, they're effectively identical performance wise

sdk/src/Temporal/Workflow.hs
```diff
@@ -746,94 +730,94 @@ defaultStartLocalActivityOptions =
...
+--   -> (ActivityArgs act :->: Workflow (Task (ActivityResult act)))
+-- startLocalActivity (activityRef -> KnownActivity codec n) opts = withWorkflowArgs @(ActivityArgs act) @(Task (ActivityResult act)) codec $ \typedPayloads -> Workflow do
+--   updateCallStack
+--   originalTime <- unWorkflow time
+--   inst <- ask
+--   runInIO <- askRunInIO
+--   let ps = fmap convertToProtoPayload typedPayloads
+--   s@(Sequence actSeq) <- nextActivitySequence
+--   resultSlot <- newIVar
+--   atomically $ modifyTVar'
```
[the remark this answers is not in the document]
jkachmar: also the module export up top

sdk/src/Temporal/Workflow.hs
```diff
@@ -646,8 +626,12 @@ startChildWorkflowFromPayloads (workflowRef -> k@(KnownWorkflow codec _)) opts p
...
                 }
 
-        atomically $ modifyTVar' inst.workflowSequenceMaps $ \seqMaps ->
-          seqMaps {childWorkflows = HashMap.insert s (SomeChildWorkflowHandle wfHandle) seqMaps.childWorkflows}
+        let anyHandle :: forall h. ChildWorkflowHandle h -> ChildWorkflowHandle Any
+            anyHandle = unsafeCoerce
+        atomically $
+          modifyTVar'
+            runtime.workflowRuntimeSequenceMaps.childWorkflows
+            (HashMap.insert s $ anyHandle wfHandle)
```
jkachmar, line 634: okay so I understand what `unsafeCoerce` is doing here, but I don't really follow what opposite side of this looks like.

you stash the `ChildWorkflowHandle Any` in the map, which gets used in `handleResolveChildWorkflowExecution` but how is the `result` type information handled here?

PR #103, opened 2025-05-06
init `tcld` at 0.40.0

jkachmar:
## description

we should upstream this to nixpkgs, but in the meantime it's handy to have this in the shell environment when interacting with Temporal Cloud


jkachmar:
this got merged to nixpkgs :)

PR #104, opened 2025-05-07
mention temporal in connection error messages

johnmastro:
STAB-848 reported that, if e.g. the temporal server isn't running, the client logs `Error: TonicTransportError(tonic::transport::Error(Transport, ConnectError(ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" }))))` without explicitly saying what it's trying to connect to.

tracing through what happens here:
- temporal's `sdk-core` [returns a `ClinitInitError`](https://github.com/temporalio/sdk-core/blob/9af3cb5b70fc67ace4db76df43b0e88f1e9bdc70/client/src/lib.rs#L310-L323), which is an enum wrapping other errors, some of which wrap yet other errors
- in our rust layer, [we log the error (this is the `Error: TonicTransportError` message users see) and then forward it on (but replacing the concrete error type with a string representation of it)](https://github.com/MercuryTechnologies/hs-temporal-sdk/blob/076393d42f54e8ca791afc5ff1bc4dae39802fbb/core/rust/src/client.rs#L214-L216)
- in our haskell layer, [we conditionally retry (based on the retry config - which is 15 in mwb's `config/settings.yml`), and then eventually throw a `ClientConnectionError` if the retries are exhausted](https://github.com/MercuryTechnologies/hs-temporal-sdk/blob/076393d42f54e8ca791afc5ff1bc4dae39802fbb/core/src/Temporal/Core/Client.hs#L258-L270)

what this looks like for a user is:
Screenshot 2025-05-07 at 12 26 19 PM

what this does is:
- remove the log message from our rust layer (which doesn't get any of our logging configuration)
- log from haskell instead
- explicitly mention that it's a temporal connection error

### DIFF
--- a/core/src/Temporal/Core/Client.hs
+++ b/core/src/Temporal/Core/Client.hs
@@ -49,6 +49,7 @@ module Temporal.Core.Client (
 import Control.Concurrent
 import Control.Exception
 import Control.Monad
+import Control.Monad.IO.Class
 import Control.Monad.Logger
 import Data.Aeson
 import Data.Aeson.TH
@@ -73,6 +74,8 @@ import System.Posix.Process
 import Temporal.Core.CTypes
 import Temporal.Internal.FFI
 import Temporal.Runtime
+import UnliftIO (MonadUnliftIO, withRunInIO)
+import qualified UnliftIO
 
 
 foreign import ccall "hs_temporal_connect_client" raw_connectClient :: Ptr Runtime -> CString -> TokioCall (CArray Word8) CoreClient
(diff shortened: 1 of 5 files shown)


core/src/Temporal/Core/Client.hs
```diff
@@ -258,13 +259,15 @@ connectClient rt conf = do
             case result of
               Left errFP -> do
                 err <- withForeignPtr errFP $ peek >=> cArrayToText
+                let err' = "Error connecting to Temporal server: " <> err
+                runStdoutLoggingT $ $(logWarn) err'
```
iand675, line 263: Instead of introducing `runStdoutLoggingT` here, I think `connectClient` should be `:: (MonadIO m, MonadLogger m)` and the logger is provided further up the stack

iand675 (approved):

PR #107, opened 2025-05-13
break out `startUpdate` vs `executeUpdate`

johnmastro:
refactor `update` into:
- `startUpdate`, which returns an `UpdateHandle` _without_ waiting for the update to complete. but note that `startUpdate` _does_ wait for the validator (if any) to complete (i.e. that the update reaches the `ACCEPTED` stage) so that we can signal an `UpdateFailed` error for validator failures rather than an "update not found" `RpcError` (which is what would happen if we tried to poll for the outcome where the validator rejected the update). this is the same thing [the typescript sdk does](https://github.com/temporalio/sdk-typescript/blob/ea28266ae12280df4cd8da6a3ff168dd6b47e1f7/packages/client/src/workflow-client.ts#L1457).
- `waitForUpdateOutcome`, which, given an `UpdateHandle`, waits for it to complete and returns the result
- `executeUpdate`, which composes those two start the update and wait for its result (the same behavior as the original `update`)

### DIFF
--- a/sdk/src/Temporal/Client.hs
+++ b/sdk/src/Temporal/Client.hs
@@ -59,7 +59,8 @@ module Temporal.Client (
   -- * Sending Updates to Workflows
   UpdateOptions (..),
   UpdateLifecycleStage (..),
-  update,
+  startUpdate,
+  executeUpdate,
 
   -- * Producing handles for existing workflows
   getHandle,
@@ -114,10 +115,10 @@ import qualified Proto.Temporal.Api.Update.V1.Message_Fields as Update
 import Proto.Temporal.Api.Workflowservice.V1.RequestResponse (
   GetWorkflowExecutionHistoryRequest,
   GetWorkflowExecutionHistoryResponse,
+  PollWorkflowExecutionUpdateRequest,
   QueryWorkflowRequest,
   QueryWorkflowResponse,
   UpdateWorkflowExecutionRequest,
-  UpdateWorkflowExecutionResponse,
  )
 import qualified Proto.Temporal.Api.Workflowservice.V1.RequestResponse_Fields as RR
 import qualified Proto.Temporal.Api.Workflowservice.V1.RequestResponse_Fields as WF
@@ -139,7 +140,7 @@ import Unsafe.Coerce
 -- WorkflowClient stuff
 
 workflowClient
-  :: MonadIO m
+  :: (MonadIO m)
   => Core.Client
   -> WorkflowClientConfig
   -> m WorkflowClient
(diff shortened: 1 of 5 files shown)


sdk/src/Temporal/Client.hs
```diff
@@ -139,7 +140,7 @@ import Unsafe.Coerce
 -- WorkflowClient stuff
 
 workflowClient
-  :: MonadIO m
+  :: (MonadIO m)
```
johnmastro, line 143: i'm not sure if the precommit hook made these formatting changes, or zed (which i took for a test drive here), or what, but i can change it back if anyone hates them

sdk/src/Temporal/Client.hs
```diff
@@ -995,14 +971,112 @@ update h@(WorkflowHandle _ _ c _ _ _) (KnownUpdate updateCodec updateName) opts
...
+  startUpdateFromPayloads wfH u opts inputs
+
+
+{- | Given an 'UpdateHandle', wait for the update operation to complete and return the result.
+
+This function will block until the update completes, and will return the result of the update
+or throw an exception if the update failed.
+-}
+waitForUpdateOutcome :: (MonadIO m) => UpdateHandle a -> m a
+waitForUpdateOutcome h = do
```
iand675, line 1033: Can we name this `waitUpdateResult` to match the `waitWorkflowResult` function?

johnmastro: just pushed a commit doing this

sdk/test/IntegrationSpec.hs
```diff
@@ -1553,6 +1553,51 @@ needsClient = do
...
+                  { updateId = "start-update-with-a-validator-that-throws"
+                  , updateHeaders = mempty
+                  }
+          ( useClient do
+              h <- C.start UpdateWithValidator "start-update-with-a-validator-that-throws" opts
+              C.startUpdate h testUpdate updateOpts 5
+            )
+            `shouldThrow` \case
+              UpdateFailure _ -> True
+              _ -> False
```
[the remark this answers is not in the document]
johnmastro: ah, right. what i really want to test here is that we threw an `UpdateFailure` and not, for example, an `RpcError`.

reading the hspec docs, it sounds like i would want to write a selector for that. is that right?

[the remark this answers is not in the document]
iand675: Well, you are testing that on line 1599. You just don't need the wildcard match to do that because you've already constrained the type by pattern matching.

sdk/src/Temporal/Client.hs
iand675: I tried zed and it did this too. I set `"format_on_save": "off"` in the config to make it a bit less aggressive about doing this. I don't particularly mind the change though

sdk/test/IntegrationSpec.hs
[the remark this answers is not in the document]
johnmastro: oh, i see. all right, just pushed a commit to remove the redundant branches

iand675 (approved):

PR #109, opened 2025-05-15
STAB-528 Decode RpcError details into nicer contents

iand675:
The current RpcError format as returned directly from Temporal's core SDK has made it unusable for engineers to better react to specific failure modes. This PR uses the protobuf `Any` mechanism to decode the current possible failure modes into ADTs

### DIFF
--- a/core/src/Temporal/Core/CTypes.hsc
+++ b/core/src/Temporal/Core/CTypes.hsc
@@ -146,8 +146,6 @@ data RpcError = RpcError
   }
   deriving (Show)
 
-instance Exception RpcError
-
 data CRPCError = CRPCError
   { code :: Word32
   , message :: CString
--- a/sdk/src/Temporal/Client.hs
+++ b/sdk/src/Temporal/Client.hs
@@ -239,10 +239,6 @@ instance HasWorkflowClient ((->) WorkflowClient) where
   askWorkflowClient = id
 
 
-throwEither :: (MonadIO m, Exception e) => IO (Either e a) -> m a
-throwEither = either throwIO pure <=< liftIO
-
-
 {- | Run a workflow, synchronously waiting for it to complete.
 
 This function will block until the workflow completes, and will return the result of the workflow
@@ -358,7 +354,7 @@ signal (WorkflowHandle _ _t c wf r _) (signalRef -> (KnownSignal sName sCodec)) 
   -- FIXME: Can we just ignore this now that it's no longer present?
   -- & WF.skipGenerateWorkflowTask .~ opts.skipGenerateWorkflowTask
   case result of
-    Left err -> throwIO err
+    Left err -> throwIO $ Temporal.Exception.coreRpcErrorToRpcError err
     Right _ -> pure ()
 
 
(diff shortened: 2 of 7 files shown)


core/src/Temporal/Core/CTypes.hsc
```diff
@@ -146,8 +146,6 @@ data RpcError = RpcError
   }
   deriving (Show)
 
-instance Exception RpcError
```
iand675, line 149: By getting rid of this instance, we can find all the callsites where it's being thrown from.

sdk/src/Temporal/Client.hs
```diff
@@ -241,10 +241,6 @@ instance HasWorkflowClient ((->) WorkflowClient) where
   askWorkflowClient = id
 
 
-throwEither :: (MonadIO m, Exception e) => IO (Either e a) -> m a
-throwEither = either throwIO pure <=< liftIO
```
iand675, line 245: I ended up expanding this into the explicit version of the same thing at callsites. It's slightly more verbose, but now that we need to perform a conversion step, it feels like readability improvements from this are a bit of a wash.

sdk/src/Temporal/Exception.hs
```diff
@@ -1,31 +1,98 @@
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 module Temporal.Exception (
-  module Temporal.Exception,
-  ActivityType (..),
-  RetryState (..),
+  -- * Common types
```
iand675, line 6: I decided to make exports explicit so we can start having things that are explicitly internal

sdk/src/Temporal/Exception.hs
```diff
@@ -563,3 +630,228 @@ data WorkflowExecutionFailedAttributes = WorkflowExecutionFailedAttributes
   , newExecutionRunId :: Maybe RunId
   }
   deriving stock (Show, Eq)
+
+data RPCStatusCode
```
iand675, line 634: Conversion of gRPC status codes to semantic equivalents

sdk/src/Temporal/Exception.hs
```diff
@@ -563,3 +630,228 @@ data WorkflowExecutionFailedAttributes = WorkflowExecutionFailedAttributes
...
+    RpcErrorWorkflowExecutionAlreadyStarted WorkflowExecutionAlreadyStartedFailure
+  | -- | The workflow is not ready to handle the operation.
+    RpcErrorWorkflowNotReady WorkflowNotReadyFailure
+  | -- | An unrecognized error type was received from the server.
+    RpcErrorUnrecognized Any.Any
+  deriving stock (Show, Eq)
+
+
+errorRegistry :: HashMap.HashMap Text (Any.Any -> Either Any.UnpackError RpcErrorDetails)
+errorRegistry = HashMap.fromList
```
iand675, line 794: The `proto-lens` library has its own concept of a registry that uses existentially quantified type class dictionary tricks, but I found that specializing our own version that provides a function for each constructor for a well-known type was a fair bit more intuitive.

jkachmar: ah yeah this is a clear improvement over what I had been thinking of, nice

jkachmar (approved):

sdk/src/Temporal/Exception.hs
```diff
@@ -563,3 +630,228 @@ data WorkflowExecutionFailedAttributes = WorkflowExecutionFailedAttributes
...
+unpackStatus s = Any.unpack $ V.head (s ^. Status.vec'details)
+
+decodeFromProof :: Message msg => Proxy msg -> BS.ByteString -> Either String msg
+decodeFromProof _ bs = decodeMessage bs
+
+-- | Detailed error information for various types of RPC errors returned by the Temporal server.
+-- This type provides specific error payloads depending on the kind of error that occurred.
+-- The Temporal API returns these details within the gRPC status, which this SDK unpacks
+-- into this more usable form.
+data RpcErrorDetails
```
jkachmar, line 753: there's no straightforward way to sanity check that all the failures exposed by the generated protobuf code, yea?

e.g. if we bump the pinned `sdk-core` commit, rerun `protogen`, and there are more failure types that have been added.

maybe there could be some sort of compile-time test that grabs all the `Type`s from each of the constructors in `RpcErrorDetails` & then compares that list w/ `discoverInstances` on all the `Message` types in scope from the `Errordetails` module

iand675: Yeah, the tricky thing with `discoverInstances` is that it's transitive, so things like `Any` will bleed into the instance list. And if you look at all of the constructors in the module using TH, sometimes they are a nested type rather than a top-level error type. We don't really want those to make it into `RpcErrorDetails`. Given those edge cases and the probably infrequent nature of additions, I feel like we're probably better off just occasionally updating this constructor by hand. Maybe we'd want a basic PR-level check that reminds people to update this if the proto module changes though?

PR #112, opened 2025-05-21
Implement some of the workflow execution listing APIs in the Temporal.Client module.

iand675:
This is to facilitate teams wanting to inspect some of their workflows from a table or something

### DIFF
--- a/sdk/src/Temporal/Client.hs
+++ b/sdk/src/Temporal/Client.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE TypeFamilyDependencies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE OverloadedLabels #-}
 {-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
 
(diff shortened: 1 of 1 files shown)


sdk/src/Temporal/Client.hs
```diff
@@ -891,25 +904,58 @@ waitResult wfId mrId (Namespace ns) = do
...
+            go c (req & #nextPageToken .~ (x ^. #nextPageToken))
 
--- countWorkflowExecutions
---   :: (MonadIO m, HasWorkflowClient m)
---   => CountWorkflowExecutionsRequest
---   -> m Int64
+countWorkflowExecutions
+  :: (MonadIO m, HasWorkflowClient m)
+  => CountWorkflowExecutionsRequest
+  -> m CountWorkflowExecutionsResponse
```
jkachmar, line 953: are there any fields here that are more useful than the `count :: Int64` one?

i see `CountWorkflowExecutionsResponse'AggregationGroup`, which i guess is useful if the count query has a group by clause?

iand675: Yeah, I guess. I don't really know what it's for, so I opted to not overly constrain the API due to lack of understanding :)

PR #113, opened 2025-05-22
avoid depending on Data.ProtoLens.Any

johnmastro:
the dependency on `Data.ProtoLens.Any` was transitively pulling in `Data.ProtoLens.Labels`, which was causing us problems via a conflict with `Data.Generic.Labels`. so, remove that dependency.

### DIFF
--- a/sdk/src/Temporal/Exception.hs
+++ b/sdk/src/Temporal/Exception.hs
@@ -1,62 +1,61 @@
 {-# LANGUAGE DuplicateRecordFields #-}
-{-# LANGUAGE TemplateHaskell #-}
 
 module Temporal.Exception (
   -- * Common types
-  ActivityType(..),
-  RetryState(..),
+  ActivityType (..),
+  RetryState (..),
 
   -- * Runtime errors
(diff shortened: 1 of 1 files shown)


sdk/src/Temporal/Exception.hs
```diff
@@ -733,23 +775,30 @@ instance Enum RPCStatusCode where
...
-unpackStatus :: Message msg => Status -> Either Any.UnpackError msg
-unpackStatus s = Any.unpack $ V.head (s ^. Status.vec'details)
+
+unpackStatus :: Message msg => Status -> Either UnpackError msg
+unpackStatus s = unpackAny $ V.head (s ^. Status.vec'details)
+
 
 decodeFromProof :: Message msg => Proxy msg -> BS.ByteString -> Either String msg
-decodeFromProof _ bs = decodeMessage bs
+decodeFromProof _ = decodeMessage
```
johnmastro, line 794: hlint wanted this

iand675 (approved):

PR #139, opened 2025-06-18
feat(sdk): add 'MockActivityEnvironment'

jkachmar:
## description

* reduce coupling between the underlying Temporal runtime & the
  'ActivityEnv' type definition
  * we don't need a handle to the worker or runtime proper; all we
    really need is the client, a handle to some function that can make
    heartbeat calls, and the worker config
  * stubbing some of this out means that we can actually mock heartbeats
    and present a more self-contained interface for testing Activities
* implement 'MockActivityEnvironment' and helper functions to define
  a reasonable default mock environment for testing
* add a few tests to validate & demonstrate some basic functionality

### DIFF
--- a/sdk/package.yaml
+++ b/sdk/package.yaml
@@ -151,5 +151,6 @@ tests:
       IntegrationSpec.NoOpWorkflow,
       IntegrationSpec.TimeoutsInWorkflows,
       IntegrationSpec.Updates,
+      MockActivityEnvironmentSpec,
       Common,
       THCompiles
--- a/sdk/src/Temporal/Activity/Definition.hs
+++ b/sdk/src/Temporal/Activity/Definition.hs
@@ -14,7 +14,7 @@ import GHC.TypeLits
 import Temporal.Activity.Types
 import Temporal.Client.Types
 import Temporal.Core.Client (Client)
-import Temporal.Core.Worker (Worker, WorkerType (Real), getWorkerClient)
+import Temporal.Core.Worker (ActivityHeartbeat, Worker, WorkerConfig, WorkerError, WorkerType (Real), getWorkerClient)
 import Temporal.Payload
 import Temporal.Workflow.Types
 import UnliftIO
@@ -57,10 +57,18 @@ data ActivityDefinition env = ActivityDefinition
 
 
 data ActivityEnv env = ActivityEnv
-  { activityWorker :: {-# UNPACK #-} !(Worker 'Real)
-  , activityInfo :: {-# UNPACK #-} !ActivityInfo
+  { activityInfo :: {-# UNPACK #-} !ActivityInfo
   , activityClientInterceptors :: {-# UNPACK #-} !ClientInterceptors
   , activityPayloadProcessor :: {-# UNPACK #-} !PayloadProcessor
+  , activityRecordHeartbeat :: ActivityHeartbeat -> IO (Either WorkerError ())
+  , activityWorkerConfig :: {-# UNPACK #-} !WorkerConfig
+  , activityWorkerClient :: Client
+  -- ^ __NOTE__: This field is /deliberately/ lazy in order to support the
+  -- 'MockActivityEnvironment' type.
+  --
+  -- When 'runMockActivity' translates the mocked environment to 'ActivityEnv',
+  -- initializes this field with a lazy exception that will only throw if the
+  -- activity-under-test (illegally) accesses the worker client.
   , activityEnv :: env
   }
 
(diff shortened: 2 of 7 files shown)


sdk/src/Temporal/Testing/MockActivityEnvironment.hs
```diff
@@ -0,0 +1,79 @@
...
+
+
+-- | FIXME: Documentation
+data MockActivityEnvironment env = MockActivityEnvironment
+  { info :: {-# UNPACK #-} !ActivityInfo
+  , payloadProcessor :: {-# UNPACK #-} !PayloadProcessor
+  , recordHeartbeat :: Worker.ActivityHeartbeat -> IO (Either Worker.WorkerError ())
+  , workerConfig :: {-# UNPACK #-} !Worker.WorkerConfig
+  , env :: env
+  }
```
jkachmar, line 26: `OverloadedRecordDot` really makes me want to just leave these fields unadorned; I feel like it should be fine to do so?

anyone who wants to use them as functions can import the module qualified, and maybe I can put that in documentation

sdk/src/Temporal/Testing/MockActivityEnvironment.hs
```diff
@@ -0,0 +1,79 @@
...
+        mempty
+        env.payloadProcessor
+        env.recordHeartbeat
+        env.workerConfig
+        undefined -- FIXME: throw a proper (impure) expression so if someone accesses it they get a reasonable error
+        env.env
+  runActivity actEnv activity
+
+-- | FIXME: Documentation.
+mkMockActivityEnvironment :: MonadIO m => env -> m (MockActivityEnvironment env)
```
jkachmar, line 45: I was originally going to just make this a pure `MockActivityEnvironment` & `unsafePerformIO` the `IORef`, but it's gotta take `env` as an argument anyway & whoever is going to be calling `runMockActivity` is going to be in `IO` (or similar) anyway.

sdk/src/Temporal/Activity/Definition.hs
```diff
@@ -57,10 +57,18 @@ data ActivityDefinition env = ActivityDefinition
...
   , activityPayloadProcessor :: {-# UNPACK #-} !PayloadProcessor
+  , activityRecordHeartbeat :: ActivityHeartbeat -> IO (Either WorkerError ())
+  , activityWorkerConfig :: {-# UNPACK #-} !WorkerConfig
+  , activityWorkerClient :: Client
+  -- ^ __NOTE__: This field is /deliberately/ lazy in order to support the
+  -- 'MockActivityEnvironment' type.
+  --
+  -- When 'runMockActivity' translates the mocked environment to 'ActivityEnv',
+  -- initializes this field with a lazy exception that will only throw if the
+  -- activity-under-test (illegally) accesses the worker client.
```
jkachmar, line 71: this is something we'll have to be cognizant of if we ever turn on `StrictData`.

if that happens, though, the test suite should fail now that we have examples that exercise `withHeartbeat` & `ActivityInfo` (both of which would force the thunk).

sdk/src/Temporal/Testing/MockActivityEnvironment.hs
```diff
@@ -0,0 +1,125 @@
...
+          , runId = "test-run"
+          , scheduleToCloseTimeout = Just $ Duration.seconds 1
+          , scheduledTime = utcToSystemTime $ UTCTime systemEpochDay 0
+          , startToCloseTimeout = Just $ Duration.seconds 1
+          , startedTime = utcToSystemTime $ UTCTime systemEpochDay 0
+          , taskToken = "test"
+          , workflowId = "test"
+          , workflowNamespace = "default"
+          , workflowType = "test"
+          }
```
jkachmar, line 102: most of these are copied from the [Python SDK's defaults for their testing environment](https://github.com/temporalio/sdk-python/blob/53d9ace634e1b25ec95d7943207714a390752fe1/temporalio/testing/_activity.py#L23-L43)

sdk/src/Temporal/Testing/MockActivityEnvironment.hs
```diff
@@ -0,0 +1,125 @@
...
+    -- NOTE: 'IllegalWorkerClientException' is thrown lazily, and will only be
+    -- surfaced to the end-user if a call is made to 'askActivityClient'.
+    actEnv =
+      ActivityEnv
+        env.info
+        mempty
+        env.payloadProcessor
+        env.recordHeartbeat
+        env.workerConfig
+        (throw IllegalWorkerClientException)
```
jkachmar, line 62: I feel like the actual best practice here would be to come up with some interface for what client functions we want to support calling from within `Activity`.

maybe a record of functions that represents a series of calls that are explicitly supported? that feels like it would be a little too inflexible though, since the whole point is to let an activity do _something_ that requires the runtime.

sdk/package.yaml
```diff
@@ -151,5 +151,6 @@ tests:
       IntegrationSpec.NoOpWorkflow,
       IntegrationSpec.TimeoutsInWorkflows,
       IntegrationSpec.Updates,
+      MockActivityEnvironmentSpec,
```
matto-mercury, line 154: is this not still going to cause hidden-module complaints from cabal?

matto-mercury: so what should I be doing in the temporal-102 ports rather than using `runMockActivity` directly?

PR #144, opened 2025-06-30
Add quasiquote-backed registration functions to make continue-as-new nicer

iand675:
[STAB-510](https://linear.app/mercury/issue/STAB-510/better-continue-as-new-support)

TL;DR– register all the things without any major tradeoffs! This lets us register all workflows and activities without declaration group shenanigans, and also lets functions reference their datatype equivalent, so that continue as new doesn't require manual instance generation any more!

Example. This works:

``` haskell
-- Register all functions at once using the simple register function
$( register
    [d|
      simpleWorkflow :: Int -> Workflow Bool
      simpleWorkflow x = pure (x == 0)


      stringWorkflow :: String -> Int -> Workflow String
      stringWorkflow s n = pure $ s ++ show n


      voidWorkflow :: Workflow ()
      voidWorkflow = provideCallStack $ do
        void $ executeChildWorkflow SimpleWorkflow defaultChildWorkflowOptions 42
        continueAsNew VoidWorkflow defaultContinueAsNewOptions


      simpleActivity :: Activity () Bool
      simpleActivity = pure True


      greetingActivity :: String -> Activity () String
      greetingActivity s = pure $ "Hello " ++ s


      customWorkflow :: Double -> Workflow Double
      customWorkflow x = pure (x * 2.0)


      customActivity :: Int -> Activity () Int
      customActivity n = pure (n + 1)
      |]
 )
```

### DIFF
--- a/sdk/src/Temporal/TH.hs
+++ b/sdk/src/Temporal/TH.hs
@@ -36,19 +36,198 @@ declarative manner. This module simplifies the process of integrating Temporal
 with Haskell applications by automating much of the boilerplate code typically
 required for setting up Workflows and Activities.
 
-Key features of this module include:
+= Registration Functions
 
-1. Automatic registration of Workflows and Activities.
-2. Type-safe configuration options for both Workflows and Activities.
-3. Generation of necessary typeclass instances for Workflow and Activity functions.
-4. Support for custom naming and aliasing of Activities.
-5. Integration with Haskell's strong type system to ensure correctness at compile-time.
+The module provides several registration functions that automatically discover and
+register Workflow and Activity functions from declaration groups:
+
+== Basic Registration
+
+@
+-- Simple registration with default options
+\$(register [d|
+  myWorkflow :: Int -> Workflow Bool
+  myWorkflow x = pure (x > 0)
+
+  myActivity :: String -> Activity () String
+  myActivity s = pure $ "Hello " ++ s
+  |])
+@
+
+== Advanced Registration with Configuration
+
+@
+-- Registration with custom configuration functions
+\$(registerWithConfig (RegistrationConfig
+  -- Custom workflow configuration
+  (\name typ -> defaultWorkflowConfig
+    { workflowConfigNameOverride = Just $ "custom-" <> Text.pack (show name)
+    , workflowConfigAliases = ["alias-" <> Text.pack (show name)]
+    })
+  -- Custom activity configuration
+  (\name typ -> defaultActivityConfig
+    { activityConfigNameOverride = Just $ "custom-" <> Text.pack (show name)
+    , activityConfigAliases = ["alias-" <> Text.pack (show name)]
+    })
+  ) [d|
+  complexWorkflow :: String -> Int -> Workflow (String, Int)
+  complexWorkflow s n = pure (s, n * 2)
+
+  complexActivity :: Bool -> Activity () Int
+  complexActivity b = pure $ if b then 1 else 0
+  |])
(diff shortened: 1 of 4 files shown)


sdk/src/Temporal/TH.hs
```diff
@@ -36,19 +36,198 @@ declarative manner. This module simplifies the process of integrating Temporal
...
+
+== Activity Configuration
+
+@
+data ActivityConfig codec = ActivityConfig
+  { activityConfigNameOverride :: Maybe Text.Text  -- ^ Override the activity name
+  , activityConfigAliases :: [Text.Text]           -- ^ Additional names for the activity
+  , activityConfigCodec :: codec                   -- ^ Codec for serialization
+  }
+@
```
jkachmar, line 118: is this redundant with the actual haddocks on these types?

sdk/src/Temporal/TH.hs
```diff
@@ -262,5 +463,466 @@ discoverDefinitions wfs acts =
...
+        , workflowConfigAliases = ["string-workflow-" <> Text.pack (show name)]
+        }
+    -- Functions taking Int get different handling
+    TH.AppT (TH.AppT TH.ArrowT (TH.ConT ''Int)) _ ->
+      defaultWorkflowConfig
+        { workflowConfigNameOverride = Just $ "int-" <> Text.pack (show name)
+        , workflowConfigAliases = ["int-workflow-" <> Text.pack (show name)]
+        }
+    -- Default configuration
+    _ -> defaultWorkflowConfig)
```
jkachmar, line 808: nit: idk if this is a good module for it, but we might want to export a pattern synonym for basic matchers that people might want to case off of.

e.g. (bad name):
```haskell
pattern Arg1 ty = TH.AppT (TH.AppT TH.ArrowT (TH.ConT ty))
```

or maybe it's just worth documenting that pattern synonyms might be helpful here with some examples and leave that up to the end user?

sdk/src/Temporal/TH.hs
```diff
@@ -262,5 +463,466 @@ discoverDefinitions wfs acts =
...
+
+  -- Configure activities based on their environment type
+  (\name typ -> case typ of
+    -- Functions with () environment get simple config
+    TH.AppT (TH.AppT (TH.ConT ''Activity) (TH.TupleT 0)) _ ->
```
jkachmar, line 813: similar thought as the pattern synonym thing above.

tuples really feel both terribly useful for quick examples and also terribly underserved in the sense that like `AppT (AppT (TH.TupleT 2) ty1) ty2` is obnoxiously tedious

sdk/test/RegisterWorkflowsSpec.hs
```diff
@@ -0,0 +1,98 @@
...
+testActivityReferences = provideCallStack do
+  _ <- executeActivity SimpleActivity (defaultStartActivityOptions $ StartToClose $ seconds 1)
+  _ <- executeActivity GreetingActivity (defaultStartActivityOptions $ StartToClose $ seconds 1) "world"
+  _ <- executeActivity CustomActivity (defaultStartActivityOptions $ StartToClose $ seconds 1) 42
+  _ <- executeActivity DynamicActivity (defaultStartActivityOptions $ StartToClose $ seconds 1) "test"
+  pure ()
+
+
+spec :: Spec
+spec = pure ()
```
jkachmar, line 98: do we want a test for the logic is implemented in some expected fashion or is successful compilation sufficient?

PR #145, opened 2025-07-02
base64 1.0 compatibility shim

jkachmar:
## description

* adds some CPP to conditionally support `base64-0.4` & `base64-1.0`
* drops `apple_sdk` + frameworks in favor of `apple-sdk`, which will become an eval warning in newer nixpkgs


flake.nix
```diff
@@ -53,8 +53,7 @@
     };
 
     overlays = {
-      temporal-bridge = import ./nix/overlays/temporal-bridge/overlay.nix;
-      temporal-test-server = import ./nix/overlays/temporal-test-server.nix;
```
jkachmar, line 57: while I'm making Nix changes, I squashed this down into a single "native" overlay that has the stuff we need to apply to top-level nixpkgs in one place

iand675 (approved):

sdk/src/Temporal/Payload.hs
```diff
@@ -441,10 +449,32 @@ data Payload = Payload
...
+
+
+-- | Internal compatibility shim to bridge support for `base64 < 1.0`.
+base64EncodeToText :: ByteString -> T.Text
+base64EncodeToText =
+#if MIN_VERSION_base64(1,0,0)
+  Base64.Types.extractBase64 . Base64.encodeBase64
+#else
+  Base64.encodeBase64
+#endif
```
jkachmar, line 477: CPP is pretty much always a little ugly but (ime) the most pleasant way to go about it is:
- import everything qualified so you can re-implement pieces of the API that may no longer be present
- be explicit about import lists even with qualification so it's extremely obvious where everything comes from
- place top-level function declarations outside of the CPP block and minimize the scope of the conditional implementation within the function as much as possible

PR #146, opened 2025-07-09
Fix fnName codegen

iand675:


sdk/src/Temporal/TH/Internal.hs
```diff
@@ -50,7 +50,12 @@ fnSingDataAndConName n
 
 makeFnDecls :: forall m. (TH.Quote m, TH.Quasi m) => TH.Name -> TH.Type -> m [TH.Dec]
 makeFnDecls n t = do
+  (TH.Module _ (TH.ModName modName)) <- TH.runQ TH.thisModule
   let dName = fnSingDataAndConName n
+      determinedName = case TH.nameModule n of
+        Nothing -> mconcat [modName, ".", TH.nameBase n]
+        Just existing -> mconcat [existing, ".", TH.nameBase n]
```
jkachmar, line 57: ah that's what the CI failure was, then?

iand675: Yeah, GHC does this annoying thing where since the functions aren't actually spliced into the module yet, the `Show` instance for `Name` chucks a wobbly by giving you an unqualified name with a suffix identifying which generated `Name` it was or something:

So, `Mercury.Workflows.GrowthInfra.DeleteAndSuppressAllUsersOnCustomerIO.deleteAndSuppressAllUsersOnCustomerIOWorkflow` would end up as `deleteAndSuppressAllUsersOnCustomerIOWorkflow_6989586621681804549` instead.

We fix that by just not using the `Show` instance.

PR #148, opened 2025-07-14
Dependency update scripting

iand675:
Gives us a single command to bump revisions. We can specify `next` to fetch the next revision for the Temporal SDK via the GitHub API (defaulting to `master`[[ as the branch). Otherwise, it'll bump to the latest revision.

image

### DIFF
--- a/scripts/update-temporal-revision.sh
+++ b/scripts/update-temporal-revision.sh
@@ -215,7 +215,7 @@ get_next_revision() {
     done
 
     if [[ $current_index -ge 0 ]]; then
-      # Found the current revision, get the prior commit (previous in the array since it's reverse chronological)
+      # Found the current revision, get the next commit (previous in the array since it's reverse chronological)
       if [[ $current_index -gt 0 ]]; then
         next_commit=$(echo "$response" | jq -r ".[$((current_index - 1))].sha")
         log_info "Found next commit: $next_commit" >&2


scripts/update-temporal-revision.sh
```diff
@@ -0,0 +1,469 @@
...
+      commit_sha=$(echo "$response" | jq -r ".[$i].sha")
+      if [[ $commit_sha == "$current_revision" ]]; then
+        current_index=$i
+        log_info "Found current revision at index $i on page $page" >&2
+        break
+      fi
+    done
+
+    if [[ $current_index -ge 0 ]]; then
+      # Found the current revision, get the prior commit (previous in the array since it's reverse chronological)
```
jkachmar, line 218:
```suggestion
      # Found the current revision, get the next commit (previous in the array since it's reverse chronological)
```
?

scripts/update-temporal-revision.sh
```diff
@@ -0,0 +1,469 @@
...
+  echo "$next_commit"
+}
+
+# Function to extract current revision from Cargo.toml
+get_current_revision() {
+  local cargo_toml="$1"
+
+  # Look for any temporal dependency and extract its revision
+  local current_revision
+  current_revision=$(grep -A 1 "temporal.*= { git = " "$cargo_toml" | grep "rev = " | head -1 | sed 's/.*rev = "\([^"]*\)".*/\1/' | tr -d '\n')
```
jkachmar, line 255: `grep` & `sed` are probably fine but if we wanted something more structured this could use [yq](https://mikefarah.gitbook.io/yq/)

PR #204, opened 2025-08-01
STAB-1051: Support Workflow ID conflict policies in the Temporal SDK

acolinhe:
### Context

Added support for WorkflowIdConflictPolicy to control how conflicts are handled when starting workflows with existing IDs.

### What changed?

- Added a new `WorkflowIdConflictPolicy` enum with four options
- Added `workflowIdConflictPolicy` field to `StartWorkflowOptions`
- Implemented conversion functions between our enum and the protobuf representation
- Added tests for each policy option

Links
* [Linear Ticket](https://linear.app/mercury/issue/STAB-1051/support-workflow-id-conflict-policies-in-the-temporal-sdk)
* [Slack Conversation](https://mercurytechnologies.slack.com/archives/C05SERU25H9/p1753199915638699)

PR #208, opened 2025-08-21
[STAB-1051]: Update conflict policy & add test case

acolinhe:
### Context

The `WorkflowIdConflictPolicy` previous change introduced a regression where workflows using `WorkflowIdReusePolicyTerminateIfRunning` began failing with:

```
Invalid WorkflowIDReusePolicy: WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING cannot be used together with a WorkflowIDConflictPolicy
```

### Root Cause

In `src/Temporal/Client.hs`, the default was set to `WorkflowIdConflictPolicyFail`. This sent a conflict policy for ALL workflows, even when not explicitly set. Since `TerminateIfRunning` is incompatible with any conflict policy, existing workflows broke.

### Solution

Changed default to `WorkflowIdConflictPolicyUnspecified`. Unspecified means "no policy set", which maintains backward compatibility.

### Testing

- Added regression test: "Default conflict policy works with TerminateIfRunning reuse policy"
- Verified test fails without fix, passes with fix
- All existing tests pass

### Links

[Linear Ticket](https://linear.app/mercury/issue/STAB-1051/support-workflow-id-conflict-policies-in-the-temporal-sdk)
[Previous PR](https://github.com/MercuryTechnologies/hs-temporal-sdk/pull/204)
[Postmortem ](https://www.notion.so/mercurytechnologies/enginc-345-temporal-workflows-using-terminate-if-running-id-reuse-policy-are-f-249951a7f21581509112cb91c77b8420?source=copy_link)

### DIFF
--- a/sdk/src/Temporal/Client.hs
+++ b/sdk/src/Temporal/Client.hs
@@ -594,7 +594,7 @@ startFromPayloads k@(KnownWorkflow codec _) wfId opts payloads = do
                 (fromMaybe WorkflowIdReusePolicyAllowDuplicateFailedOnly opts'.workflowIdReusePolicy)
             & WF.workflowIdConflictPolicy
               .~ workflowIdConflictPolicyToProto
-                (fromMaybe WorkflowIdConflictPolicyFail opts'.workflowIdConflictPolicy)
+                (fromMaybe WorkflowIdConflictPolicyUnspecified opts'.workflowIdConflictPolicy)
             & WF.maybe'retryPolicy .~ (retryPolicyToProto <$> opts'.retryPolicy)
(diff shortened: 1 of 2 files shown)


acolinhe:
Below are screenshots before/after making the changes to `sdk/src/Temporal/Client.hs` w/ the included test case addition. We see that before the change, the test case rightfully fails. 

**Before**
Screenshot 2025-08-21 at 12 08 10 PM

**After**
Screenshot 2025-08-21 at 12 09 25 PM

johnmastro (approved):

PR #212, opened 2025-08-29
Adapt to API changes of recent packages

avdv:
This PR allows building with more recent dependencies, namely `text`, `random` and `Cabal`.


sdk/src/Temporal/Exception.hs
```diff
@@ -72,7 +72,7 @@ import qualified Data.HashMap.Strict as HashMap
 import Data.Int
 import Data.ProtoLens (Message (..), decodeMessage, decodeMessageOrDie)
 import Data.ProtoLens.Field (field)
-import Data.Text
+import Data.Text (Text, pack, breakOnEnd)
```
avdv, line 75: Usually, it's recommended to import `Data.Text` qualified but this seemed easy enough. Alternatively, I could also do that if you want.

iand675:
Apologies, I didn't see this PR! Thanks!

PR #216, opened 2025-09-05
add inbound execute workflow wrapper

acolinhe:
## Context

This adds a new inbound hook via `executeWorkflowWrapped`​ that lets the `WorkflowInboundInterceptor` provide a small wrapper `(Workflow Payload -> Workflow Payload)`​ applied right after decoding and before execution. The worker can now call this wrapped entry point that passes id by default, so existing behavior is unchanged. OpenTelemetry is updated to mirror the same logic for the wrapped path.

This enables downstream use cases, like accessing runId, without changing workflow code. A follow‑up PR will add upsertMemo.

Local build & test run of changes:

Test Pass Screenshot


acolinhe:
* **#216** <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/hs-temporal-sdk/216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/MercuryTechnologies/hs-temporal-sdk/216?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

sdk/src/Temporal/Contrib/OpenTelemetry.hs
```diff
@@ -201,6 +201,73 @@ makeOpenTelemetryInterceptor = do
                       recordException span mempty Nothing e
                     _ -> pure ()
                   pure execution
+            , -- TODO: Consider deduping common span setup with executeWorkflow to avoid drift
```
acolinhe, line 204: Mirroring the existing tracing logic for the wrapped path to keep spans/attributes identical when the wrapper is used.

sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -848,6 +848,15 @@ data WorkflowInboundInterceptor = WorkflowInboundInterceptor
       :: ExecuteWorkflowInput
       -> (ExecuteWorkflowInput -> IO (WorkflowExitVariant Payload))
       -> IO (WorkflowExitVariant Payload)
+  , -- Same semantics as executeWorkflow, but with a start-time wrapper
+    executeWorkflowWrapped
```
acolinhe, line 852: Add `executeWorkflowWrapped` so inbound interceptor can supply a start time Workflow wrapper while preserving the old entry‑point for backwards compatibility.

sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -848,6 +848,15 @@ data WorkflowInboundInterceptor = WorkflowInboundInterceptor
...
+      :: ExecuteWorkflowInput
+      -> (Workflow Payload -> Workflow Payload)
+      -> ( ExecuteWorkflowInput
+           -> (Workflow Payload -> Workflow Payload)
```
acolinhe, line 856: Wrapper injected here

acolinhe:
Discussed w/ @matto-mercury that I didn't add test cases in this pr because there aren't direct tests for `executeWorkflow` and `applyStartWorkflow` , that I see in the temporal-sdk. There are indirect tests in [sdk/test/IntegrationSpec.hs](https://github.com/MercuryTechnologies/hs-temporal-sdk/blob/0aafae992130432f2a8e3c99780fbcae1758ad55/sdk/test/IntegrationSpec.hs#L602), which from my understanding have the same testing path as these changes.

I'll also be adding tests in configurations that use this wrapper in the follow-up `upsertMemo` pr

sdk/src/Temporal/Contrib/OpenTelemetry.hs
```diff
@@ -201,6 +201,7 @@ makeOpenTelemetryInterceptor = do
                       recordException span mempty Nothing e
                     _ -> pure ()
                   pure execution
+            , startExecuteWorkflowHook = const id
```
acolinhe, line 204: default const id so that this `executeworkflow` is backwards compatible.

PR #219, opened 2025-09-11
add upsertMemo

acolinhe:
### Context

Add the ability to update workflow memo values during workflow execution.

### What changed?

- Add a new `upsertMemo` function to the `Temporal.Workflow` module that allows workflows to update their memo by merging new values with existing ones
- Export the new function in the module header
- Add integration tests to verify:
    - Reading memo values set at workflow start
    - Upserting new memo values during workflow execution
    - Multiple upsert operations working correctly with proper overwriting behavior

### Why make this change?

This change is being made to upsert human-readable memo fields during a workflow’s lifetime (e.g., add breadcrumbs/url links after start) to improve observability.

### Local Testing
Compiled and run locally in MWB to confirm `upsertMemo` builds cleanly and behaves as expected.

### Links
* [Ticket](https://linear.app/mercury/issue/STAB-1172/add-upsertmemo-to-temporal-sdk)
* [Write-up](https://docs.google.com/document/d/1BFR8uDA5f4b-vFldwHfF9vaU1hcpATjD5Ol9UslWb1s/edit?usp=sharing)
* [Slack Discussion](https://mercurytechnologies.slack.com/archives/C04N6SQH1FT/p1757623671528049)

### DIFF
--- a/sdk/src/Temporal/Client.hs
+++ b/sdk/src/Temporal/Client.hs
@@ -581,6 +581,7 @@ startFromPayloads k@(KnownWorkflow codec _) wfId opts payloads = do
     searchAttrs <- searchAttributesToProto opts'.searchAttributes
     payloads'' <- processorEncodePayloads c.clientConfig.payloadProcessor payloads'
     hdrs <- processorEncodePayloads c.clientConfig.payloadProcessor opts'.headers
+    memo' <- processorEncodePayloads c.clientConfig.payloadProcessor opts'.memo
     let tq = rawTaskQueue opts'.taskQueue
         req =
           defMessage
@@ -608,7 +609,7 @@ startFromPayloads k@(KnownWorkflow codec _) wfId opts payloads = do
                 (fromMaybe WorkflowIdConflictPolicyUnspecified opts'.workflowIdConflictPolicy)
             & WF.maybe'retryPolicy .~ (retryPolicyToProto <$> opts'.retryPolicy)
             & WF.cronSchedule .~ fromMaybe "" opts'.cronSchedule
-            & WF.memo .~ convertToProtoMemo opts'.memo
+            & WF.memo .~ convertToProtoMemo memo'
             & WF.searchAttributes .~ (defMessage & Common.indexedFields .~ searchAttrs)
             --     TODO Not sure how to use these yet
             & WF.header .~ headerToProto (fmap convertToProtoPayload hdrs)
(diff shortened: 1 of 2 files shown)


sdk/test/IntegrationSpec.hs
```diff
@@ -1932,6 +1932,73 @@ needsClient = do
...
+            expectedB = encodeJSON (toJSON ("two" :: Text))
+            expectedC = encodeJSON (toJSON True)
+            expected =
+              Map.fromList
+                [ ("a", dec p1)
+                , ("b", expectedB)
+                , ("c", expectedC)
+                ]
+        m `shouldBe` expected
+    specify "multiple upserts work" $ \TestEnv {..} -> do
```
acolinhe, line 1982: Verifies later calls take precedence for memo behavior

sdk/test/IntegrationSpec.hs
```diff
@@ -1932,6 +1932,73 @@ needsClient = do
         (Temporal.Client.TestService.unlockTimeSkippingWithSleep (C.clientCore (C.workflowHandleClient h)) (seconds 10))
           `shouldThrow` \case
             Temporal.Client.TestService.TimeSkippingNotSupported -> True
+  describe "Memo" $ do
+    specify "able to read memo set at start" $ \TestEnv {..} -> do
```
acolinhe, line 1936: Here, we're verifying that we're able to see what's inside of memo at the start of a workflow.

sdk/test/IntegrationSpec.hs
```diff
@@ -1932,6 +1932,73 @@ needsClient = do
...
+            baseConf
+      withWorker conf $ do
+        p1 <- encode JSON ("v1" :: Text)
+        p2 <- encode JSON (1 :: Int)
+        let opts =
+              (C.startWorkflowOptions taskQueue)
+                { C.workflowIdReusePolicy = Just W.WorkflowIdReusePolicyAllowDuplicate
+                , C.memo = Map.fromList [("a", p1), ("b", p2)]
+                }
+        let dec (Payload bs meta) = Payload (BS.map (\x -> x - 1) bs) meta
```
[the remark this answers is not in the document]
iand675: I'd expect that the client should be able to decode details as well, transparently, so I think I agree with John that this isn't quite the right behaviour here.

sdk/src/Temporal/Client.hs
```diff
@@ -581,6 +581,7 @@ startFromPayloads k@(KnownWorkflow codec _) wfId opts payloads = do
     searchAttrs <- searchAttributesToProto opts'.searchAttributes
     payloads'' <- processorEncodePayloads c.clientConfig.payloadProcessor payloads'
     hdrs <- processorEncodePayloads c.clientConfig.payloadProcessor opts'.headers
+    memo' <- processorEncodePayloads c.clientConfig.payloadProcessor opts'.memo
```
acolinhe, line 584: We're using the client payload processor to do the memo encoding

PR #222, opened 2025-09-18
add signalWithStartFromPayloads

johnmastro:
add `signalWithStartFromPayloads` (by factoring it out of `signalWithStart`)

### DIFF
--- a/sdk/src/Temporal/Client.hs
+++ b/sdk/src/Temporal/Client.hs
@@ -108,7 +108,6 @@ import Data.Map (Map)
 import Data.Maybe
 import Data.ProtoLens.Field
 import Data.ProtoLens.Message
-import Data.Proxy
 import Data.Text (Text)
 import Data.Typeable
 import qualified Data.UUID as UUID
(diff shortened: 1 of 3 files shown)


sdk/src/Temporal/Client.hs
```diff
@@ -664,6 +657,97 @@ start (workflowRef -> k@(KnownWorkflow codec _)) wfId opts = withArgs @(Workflow
...
+            , workflowHandleRunId = Just (RunId $ swer ^. WF.runId)
+            , -- We don't know if this handle represents a new workflow or an already-running one, so we don't
+              -- know if the current run ID is for its first execution or not
+              workflowHandleFirstExecutionRunId = Nothing
+            }
+  pure $
+    wfH
+      { workflowHandleReadResult = \a ->
+          workflowHandleReadResult wfH a >>= \b -> do
+            result <- decode codec =<< either (throwIO . ValueError) pure =<< payloadProcessorDecode c.clientConfig.payloadProcessor b
```
johnmastro, line 746: this function is mostly code motion from `signalWithStart`, but this line is different and fixes a bug in the existing implementation, where we're currently not calling `payloadProcessorDecode`

iand675 (approved):

PR #226, opened 2025-09-29
Support catching ApplicationFailure in Workflows for simpler cases

iand675:
I wrote some code the other day where I tried catching an `ApplicationFailure`, but the exception thrown in the workflow was an `ActivityFailure`. Since the application failure is a field within the activity failure, this seems like a nice quality of life dynamic cast to support.

### DIFF
--- a/sdk/src/Temporal/Exception.hs
+++ b/sdk/src/Temporal/Exception.hs
@@ -404,7 +404,9 @@ applicationFailureToFailureProto appFailure =
          )
 
 
-instance Exception ApplicationFailure
+instance Exception ApplicationFailure where
+  fromException (SomeException e) =
+    cast e <|> (cause <$> cast e)
 
 
 class ToApplicationFailure e where


johnmastro (approved):

sdk/src/Temporal/Exception.hs
```diff
@@ -404,7 +404,9 @@ applicationFailureToFailureProto appFailure =
          )
 
 
-instance Exception ApplicationFailure
+instance Exception ApplicationFailure where
+  fromException (SomeException e) =
+    cast e <|> (cause <$> cast e)
```
jkachmar, line 409: `cause :: ActivityFailure -> Maybe ApplicationFailure` makes the second `cast` infer the right type & `(<|>)` peels one layer off the `Maybe (Maybe ApplicationFailure)`; just writing that out to make sense of it to myself.

PR #231, opened 2025-10-30
Eliminate all panics from Rust FFI code

iand675:
I've been bothered for a while that we're not necessarily handling failures in a way that Haskell can gracefully deal with in order to report things to sentry/whatever else. So, this does a pass to get rid of any spots under our control where Rust can chuck a wobbly.

### DIFF
--- a/codec-encryption/temporal-codec-encryption.cabal
+++ b/codec-encryption/temporal-codec-encryption.cabal
@@ -5,7 +5,7 @@ cabal-version: 3.0
 -- see: https://github.com/sol/hpack
 
 name:           temporal-codec-encryption
-version:        2025.10.1.0
+version:        2025.10.2.0
 category:       Concurrency
 author:         Ian Duncan
 maintainer:     temporal@mercury.com
--- a/core/rust/src/client.rs
+++ b/core/rust/src/client.rs
@@ -46,10 +46,12 @@ struct ClientRetryConfig {
 
 fn client_config_to_options(
     client_config: ClientConfig,
-) -> Result<ClientOptions, ClientOptionsBuilderError> {
+) -> Result<ClientOptions, String> {
     let mut defaults = ClientOptionsBuilder::default();
+    let target_url = Url::parse(&client_config.target_url)
+        .map_err(|e| format!("Invalid URL: {}", e))?;
     let mut options_builder = defaults
-        .target_url(Url::parse(&client_config.target_url).unwrap())
+        .target_url(target_url)
         .client_name(client_config.client_name)
         .client_version(client_config.client_version)
         .identity(client_config.identity);
@@ -88,7 +90,7 @@ fn client_config_to_options(
         options_builder = options_builder.api_key(client_config.api_key)
     }
 
-    options_builder.build()
+    options_builder.build().map_err(|e| format!("Failed to build client options: {}", e))
 }
 
 #[repr(C)]
@@ -153,7 +155,10 @@ impl From<&RpcCall> for TemporalCall {
             req: unsafe {
                 let req_array = rpc_call.req;
                 let rust_vec = (*req_array).as_rust();
-                rust_vec.unwrap().clone()
+                rust_vec.unwrap_or_else(|e| {
+                    eprintln!("Failed to convert RPC request: {:?}", e);
+                    vec![]
+                }).clone()
             },
             retry: rpc_call.retry,
             metadata: unsafe { convert_hashmap(rpc_call.metadata) },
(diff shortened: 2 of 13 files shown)


jkachmar:
okay i rebased this against `main` and removed a bunch of `eprintln` calls that were superfluous as they just repeat the error that we pass back across the FFI boundary; in those cases the caller can decide whether they want to do something with it.

codec-encryption/temporal-codec-encryption.cabal
```diff
@@ -5,7 +5,7 @@ cabal-version: 3.0
 -- see: https://github.com/sol/hpack
 
 name:           temporal-codec-encryption
-version:        2025.10.1.0
+version:        2025.10.2.0
```
jkachmar, line 8: oh right this should probably be 2026 or something

core/rust/src/client.rs
```diff
@@ -322,13 +363,19 @@ where
     match res {
         Ok(resp) => Ok(resp.get_ref().encode_to_vec()),
         Err(err) => {
-            eprintln!("Error: {:?}", err);
```
jkachmar, line 325: we're already successfully passing the client error struct back across the FFI, we don't need to _also_ print this out as an error call

codec-encryption/temporal-codec-encryption.cabal
andrewsmith: Do we normally bump versions in the same PR as a code change? Are we doing this because we've already shipped `2025.10.1.0` and need to update our "next" version?

core/rust/src/client.rs
```diff
@@ -153,7 +155,10 @@ impl From<&RpcCall> for TemporalCall {
             req: unsafe {
                 let req_array = rpc_call.req;
                 let rust_vec = (*req_array).as_rust();
-                rust_vec.unwrap().clone()
+                rust_vec.unwrap_or_else(|e| {
+                    eprintln!("Failed to convert Temporal client RPC request: {:?}", e);
+                    vec![]
+                }).clone()
```
andrewsmith, line 161: I don't know enough context, but does this potentially sweep the error under the rug? We hand back an empty `Vec`. How is that used later? Is there anything that expects that to be non-empty?

It seems like the case where `as_rust` could fail is if the CArray is a null pointer.

But for safety's sake, why wouldn't we do `TryFrom` and handle the error case closer to where TemporalCall is needed?

core/rust/src/client.rs
```diff
@@ -196,11 +201,20 @@ pub fn connect_client(
...
+
     runtime_ref
         .runtime
         .future_result_into_hs(hs_callback, async move {
+            let opts = opts_result.map_err(|e| {
+                let err_message = format!("Invalid client options: {}", e).into_bytes();
+                CArray::c_repr_of(err_message).unwrap_or_else(|conv_err| {
+                    eprintln!("Failed to convert Temporal client error message: {:?}", conv_err);
+                    CArray { data_ptr: std::ptr::null_mut(), size: 0 }
+                })
```
andrewsmith, line 215: This feels like one we should just `unwrap`, as there is no conversion of the elements (`u8`) that can fail when `c_repr_of` runs over them.

core/rust/src/client.rs
```diff
@@ -212,7 +226,10 @@ pub fn connect_client(
                 }),
                 Err(e) => {
                     let err_message = e.to_string().into_bytes();
-                    Err(CArray::c_repr_of(err_message).unwrap())
+                    Err(CArray::c_repr_of(err_message).unwrap_or_else(|conv_err| {
+                        eprintln!("Failed to convert Temporal client error message: {:?}", conv_err);
+                        CArray { data_ptr: std::ptr::null_mut(), size: 0 }
+                    }))
```
andrewsmith, line 232: Same here, this feels overly defensive.

And I almost worry _more_ about returning a null pointer! (Without more context of where this is called on the Haskell side)

core/rust/src/client.rs
andrewsmith, line 1: With the initial ones I am seeing in this file, I'm getting the feeling reading this that we might be jumping out of the frying pan into the fire.

At least with an `unwrap()` you are going to see a panic with a line number. You know something bad happened and where. And the program isn't going to continue. Now with potentially returning null pointers, that has to be checked on the calling side everywhere and handled properly.

Do we have confidence we are doing that properly? Am I misunderstanding some bigger safety net?

core/rust/src/client.rs
```diff
@@ -233,7 +250,22 @@ pub unsafe extern "C" fn hs_temporal_connect_client(
 ) {
     let runtime_ref = unsafe { &*runtime_ref };
     let config_json = unsafe { CStr::from_ptr(config_json) };
-    let config: ClientConfig = serde_json::from_slice(config_json.to_bytes()).unwrap();
+    let config: ClientConfig = match serde_json::from_slice(config_json.to_bytes()) {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            let err_message = format!("Failed to parse client config: {}", e).into_bytes();
+            let err_array = CArray::c_repr_of(err_message).unwrap_or_else(|conv_err| {
+                eprintln!("Failed to convert Temporal client error message: {:?}", conv_err);
```
andrewsmith, line 258: I don't see how this one would fail either.

core/rust/src/client.rs
andrewsmith: This feels like the real pertinent change with respect to the logging spam people have been talking about. Could we take this as a separate change first? I'm struggling to see how removing this would be bad separate from the other panic conversions.

PR #294, opened 2026-02-18
[wip] STAB-1400: reproduce fetchHistory issue

jkachmar:
## description

@Deniz-Jasa & i paired a bit on this, and it looks like there's some nuance to the problem, but this reproduces the original problem & demonstrates that it's up to pagination not working.

### notes

fun fact[^1]: Temporal advises that page size limits are suggestions, not concrete guarantees. it turns out that concurrent execution will result in all history events returning in a single batch.

that is, if you modify the test workflow in this PR to call `Workflow.forConcurrently_` then the test will pass! 

[^1]: actually not fun! very annoying!!

### DIFF
--- a/sdk/package.yaml
+++ b/sdk/package.yaml
@@ -153,6 +153,7 @@ tests:
       AsyncCompletionSpec,
       DurationSpec,
       IntegrationSpec,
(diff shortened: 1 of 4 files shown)


jkachmar:
resolved in #403

PR #359, opened 2026-04-03
remove extraneous 'eprintln!' calls

jkachmar:
## description

these calls can be useful for debugging in development, but in all cases they were redundant with respect to the structured error information passed across the FFI boundary

### DIFF
--- a/core/rust/src/client.rs
+++ b/core/rust/src/client.rs
@@ -322,7 +322,6 @@ where
     match res {
         Ok(resp) => Ok(resp.get_ref().encode_to_vec()),
         Err(err) => {
-            eprintln!("Error: {:?}", err);
             Err(CRPCError::c_repr_of(RPCError {
                 code: err.code() as u32,
                 message: err.message().to_owned(),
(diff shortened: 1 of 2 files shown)


core/rust/src/client.rs
```diff
@@ -321,14 +321,11 @@ where
...
-                details: err.details().into(),
-            })
-            .unwrap())
-        }
+        Err(err) => Err(CRPCError::c_repr_of(RPCError {
+            code: err.code() as u32,
+            message: err.message().to_owned(),
+            details: err.details().into(),
+        })
+        .unwrap()),
```
azuzunaga: All I know of Rust is that this caused a Cloudflare outage.

azuzunaga (approved):
Looks good. Verified unrelated diff was just formatting changes.

FYI diff was kinda noisy, probably an autoformatter making some more line breaks and ordering stuff around. It might be worth running it through the whole codebase so we don't have to run into this again.

jkachmar:
oh whoa, i guess that was rustfmt, lemme push up a commit without those...

core/rust/src/client.rs
jkachmar: ah yeah, so we have a branch that removes a bunch of the `panic`s but there's a broader design discussion worth having around what the appropriate place for panics to live should be.

not _all_ of these should be panic calls probably, but in some cases they _should_ be because we want certain classes of failure to signal that something unsound has happened and the program should exit immediately.

azuzunaga (approved):
Image

PR #360, opened 2026-04-08
add headers-based propagator for W3C baggage

jkachmar:
## description

just what it says: this gives us the ability to propagate baggage via Temporal headers as well as tracestate/tracecontext.

### notes

i wonder if we should provide some configuration options for filtering items out? honestly it might just be reasonable to put this on the caller, and say that they're responsible for making sure that the active `Context` doesn't have baggage they don't want to propagate.

only real risk i can think of is that someone puts something excessively large in there & it exceeds the size limits Temporal imposes upon us.


matto-mercury (approved):
just pattern-matching on the propagator pattern but lgtm

PR #368, opened 2026-04-22
add a watchdog timer to the activity shutdown path

jkachmar:
## description

activities can provide pathological behavior, including loops that swallow the async exceptions we depend on for task cancellation.

we should provide a watchdog that can ensure the Temporal runtime loop will terminate, even if that termination is unclean, in accordance with a user's requested fatal timeout period.

### DIFF
--- a/sdk/src/Temporal/Activity/Worker.hs
+++ b/sdk/src/Temporal/Activity/Worker.hs
@@ -12,9 +12,11 @@ import Control.Monad.Reader
 import Data.Bifunctor
 import Data.HashMap.Strict (HashMap)
 import qualified Data.HashMap.Strict as HashMap
+import Data.Maybe (catMaybes)
 import Data.ProtoLens
 import Data.Text (Text)
 import qualified Data.Text as T
+import GHC.Conc.Sync (threadLabel)
 import Lens.Family2
 import qualified ListT
 import qualified Proto.Temporal.Api.Common.V1.Message_Fields as P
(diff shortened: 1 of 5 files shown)


9999years (approved):
Wow!!! Great find!!!

sdk/src/Temporal/Activity/Worker.hs
```diff
@@ -54,12 +56,25 @@ data ActivityWorker env = ActivityWorker
...
 notifyShutdown :: MonadUnliftIO m => ActivityWorker env -> m ()
 notifyShutdown worker = do
-  let shutdownGracePeriod = fromIntegral (Core.gracefulShutdownPeriodMillis $ Core.getWorkerConfig worker.workerCore)
-  -- TODO logging here
-  when (shutdownGracePeriod > 0) $ do
+  let cfg = Core.getWorkerConfig worker.workerCore
+      shutdownGracePeriod = fromIntegral (Core.gracefulShutdownPeriodMillis cfg)
+      fatalShutdownPeriod = Core.fatalShutdownPeriodMillis cfg
+  when (shutdownGracePeriod > 0) $
     threadDelay (shutdownGracePeriod * 1000)
```
andrewsmith, line 63: This is an aside from the current change, but does this `threadDelay` imply that the overall worker shutdown will be _at least_ `gracefulShutdownPeriodMillis` long? What if the activities conclude faster than that?

andrewsmith (approved):
LGTM

PR #374, opened 2026-05-05
chore: update 'sdk-rust'

jkachmar:
## description

* temporalio/sdk-rust@9a87ebf7 -> temporalio/sdk-rust@43d582b
* updates to `client.rs`, `runtime.rs`, & `worker.rs` in response to upstream changes

### notes

big change is that upstream switched to using [`bon`](https://bon-rs.com/) for deriving builders, which means a few things:
1. some of the builders that were fallible before are no longer
2. we can't partially construct a builder and pass it around as data b/c `bon` uses trait bounds to perform completeness checks

(1) is nice, (2) is annoying but ultimately not _really_ a problem to work around

### DIFF
--- a/core/rust/Cargo.nix
+++ b/core/rust/Cargo.nix
@@ -463,6 +463,73 @@ rec {
         };
         resolvedDefaultFeatures = [ "std" ];
       };
+      "bon" = rec {
+        crateName = "bon";
+        version = "3.9.1";
+        edition = "2021";
+        sha256 = "1zmj96nj080arzvy35h3wxza2sygp7gi1hsk6djywxh6an9bwzgl";
+        dependencies = [
+          {
+            name = "bon-macros";
+            packageId = "bon-macros";
+          }
+          {
+            name = "rustversion";
+            packageId = "rustversion";
+          }
+        ];
+        features = {
+          "alloc" = [ "bon-macros/alloc" ];
+          "default" = [ "std" ];
+          "experimental-generics-setters" = [ "bon-macros/experimental-generics-setters" ];
+          "experimental-overwritable" = [ "bon-macros/experimental-overwritable" ];
+          "implied-bounds" = [ "bon-macros/implied-bounds" ];
+          "std" = [ "bon-macros/std" "alloc" ];
+        };
+        resolvedDefaultFeatures = [ "alloc" "default" "implied-bounds" "std" ];
+      };
+      "bon-macros" = rec {
+        crateName = "bon-macros";
+        version = "3.9.1";
+        edition = "2021";
+        sha256 = "0z66ygzjyr4ivp3mzn2y98yhs5yh2qnri7f2f99jvd7fd88x76si";
+        procMacro = true;
+        libName = "bon_macros";
+        dependencies = [
+          {
+            name = "darling";
+            packageId = "darling";
+          }
+          {
+            name = "ident_case";
+            packageId = "ident_case";
+          }
+          {
+            name = "prettyplease";
+            packageId = "prettyplease";
+          }
+          {
+            name = "proc-macro2";
+            packageId = "proc-macro2";
+          }
+          {
+            name = "quote";
+            packageId = "quote";
+          }
+          {
+            name = "rustversion";
+            packageId = "rustversion";
+          }
+          {
+            name = "syn";
+            packageId = "syn 2.0.117";
+            features = [ "full" "visit-mut" "visit" ];
+          }
+        ];
+        features = {
+        };
+        resolvedDefaultFeatures = [ "alloc" "default" "implied-bounds" "std" ];
+      };
       "bumpalo" = rec {
         crateName = "bumpalo";
         version = "3.20.2";
@@ -749,9 +816,9 @@ rec {
       };
       "darling" = rec {
         crateName = "darling";
-        version = "0.20.11";
+        version = "0.23.0";
         edition = "2021";
-        sha256 = "1vmlphlrlw4f50z16p4bc9p5qwdni1ba95qmxfrrmzs6dh8lczzw";
+        sha256 = "179fj6p6ajw4dnkrik51wjhifxwy02x5zhligyymcb905zd17bi5";
         authors = [
           "Ted Driggs <ted.driggs@outlook.com>"
         ];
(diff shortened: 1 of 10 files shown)


core/rust/src/client.rs
```diff
@@ -44,51 +43,62 @@ struct ClientRetryConfig {
...
-    client_config: ClientConfig,
-) -> Result<ClientOptions, ClientOptionsBuilderError> {
-    let mut defaults = ClientOptionsBuilder::default();
-    let mut options_builder = defaults
-        .target_url(Url::parse(&client_config.target_url).unwrap())
-        .client_name(client_config.client_name)
-        .client_version(client_config.client_version)
-        .identity(client_config.identity);
+impl TryFrom<ClientConfig> for ClientOptions {
+    type Error = anyhow::Error;
```
jkachmar, line 47: the concrete error types are gone from upstream, so we might as well add `anyhow` & start using it

jkachmar: also it feels a little nicer to make these `TryFrom` impls rather than their own top-level functions

core/rust/src/client.rs
```diff
@@ -44,51 +43,62 @@ struct ClientRetryConfig {
...
                         client_cert,
                         client_private_key,
                     })
                 }
-                _ => None,
+                _ => {
+                    return Err(anyhow::anyhow!(
+                        "Must have both client cert and private key or neither"
+                    ));
+                }
```
jkachmar, line 85: providing one but not the other is an error case

core/rust/src/worker.rs
```diff
@@ -62,87 +62,106 @@ pub struct TunerConfig {
...
-            message: format!("Failed building tuner: {}", err),
-        })?;
+        let maybe_workflow_slot_opts =
+            if let Some(ref ssc @ SlotSupplierConfig::ResourceBased { .. }) =
+                cfg.workflow_slot_supplier
+            {
+                Some(ssc.into())
+            } else {
+                None
+            };
```
jkachmar: surely there's a better way to do this...?

core/rust/src/worker.rs
```diff
@@ -373,27 +399,52 @@ pub struct WorkerConfig {
...
+                    conf.max_outstanding_local_activities,
+                )));
+                if let Some(m) = conf.max_outstanding_nexus_tasks {
+                    builder.nexus_slot_supplier(Arc::new(FixedSizeSlotSupplier::new(m)));
+                };
+                Ok(builder.build())
+            }
+        }
+    }
+}
```
jkachmar, line 409: `bon` builders make it annoying to choose between setting a `tuner` or calling the `max_outstanding*` fields on the Temporal `WorkerConfig`, so instead we _only_ construct the `tuner` and treat the case where no `TunerConfig` is explicitly supplied as one where we build our own `tuner` where each of the task types gets `FixedSizeSlotSupplier`s set to the raw numeric values passed in through our config.

cf. https://github.com/temporalio/sdk-rust/blob/a22517e43db15e2a7592f376775a67951c116d0d/crates/sdk-core/src/worker/tuner.rs#L220-L236

core/rust/src/worker.rs
```diff
@@ -373,27 +399,52 @@ pub struct WorkerConfig {
...
+
 impl TryFrom<WorkerConfig> for temporalio_sdk_core::WorkerConfig {
     type Error = WorkerError;
 
     fn try_from(conf: WorkerConfig) -> Result<Self, WorkerError> {
-        let mut builder = temporalio_sdk_core::WorkerConfigBuilder::default();
-        builder
+        let converted_tuner: TunerHolder = (&conf).try_into()?;
```
jkachmar, line 415: try to produce a `TunerHolder` from our `WorkerConfig`; if `conf.tuner: Nothing`, then we produce one by explicitly building from `max_outstanding*` config fields.

core/rust/src/worker.rs
```diff
@@ -424,35 +475,29 @@ impl TryFrom<WorkerConfig> for temporalio_sdk_core::WorkerConfig {
...
-            message: format!("{}", err),
-        })
+            ))
+            // FIXME: Implement 'WorkerTaskTypes' as an FFI type
+            .task_types(temporalio_common::worker::WorkerTaskTypes {
+                enable_workflows: true,
+                enable_local_activities: true,
+                enable_remote_activities: !conf.no_remote_activities,
+                enable_nexus: true,
+            })
```
jkachmar, line 469: we should add a `WorkerTaskTypes` config struct that gets unmarshalled into this but that's _also_ going to involve plumbing the config options up through our own Worker so we can selectively start up the different worker loops.

for now, everything set to `true` except for `enable_remote_activities` (which is conditional on `no_remote_activities`) should do the trick.

core/rust/src/worker.rs
jkachmar: lol there absolutely was

core/src/Temporal/Core/Worker.hs
```diff
@@ -438,6 +438,7 @@ data TunerConfig = TunerConfig
   { workflowSlotSupplier :: Maybe SlotSupplierConfig
   , activitySlotSupplier :: Maybe SlotSupplierConfig
   , localActivitySlotSupplier :: Maybe SlotSupplierConfig
+  , nexusSlotSupplier :: Maybe SlotSupplierConfig
```
jkachmar, line 441: nexus tuner config options weren't exposed before but there's no reason not to just pass them along through

sdk/src/Temporal/Worker.hs
```diff
@@ -369,6 +371,18 @@ setMaxOutstandingLocalActivities n = modifyCore $ \conf ->
...
+  conf
+    { Core.maxOutstandingNexusTasks = Just n
+    }
+
+unsetMaxOutstandingNexusTasks :: ConfigM actEnv ()
+unsetMaxOutstandingNexusTasks = modifyCore $ \conf ->
+  conf
+    { Core.maxOutstandingNexusTasks = Nothing
+    }
+
```
jkachmar, line 385: since this is optional we should provide an `unset` option

core/rust/Cargo.nix
```diff
@@ -8269,6 +8251,7 @@ rec {
           }
         ];
         features = {
+          "antithesis_assertions" = [ "dep:antithesis_sdk" ];
```
andrewsmith, line 8254: 👀

core/rust/src/worker.rs
andrewsmith: I was concerned about the notes in the [worker performance](https://docs.temporal.io/develop/worker-performance) doc that say:
```
Worker tuners supersede the existing maxConcurrentXXXTask style Worker options. Using both styles will cause an error at Worker initialization time.
```

The Rust code in hs-temporal-sdk doesn't seem like it handles that case, but then I found it is handled in the core here: https://github.com/temporalio/sdk-rust/blob/43d582b57fbc63d858a35f11ebfee193bda00834/crates/common/src/worker.rs#L335

core/rust/src/worker.rs
andrewsmith: That seems fine, but it would be nice to have a tracking issue in this repo for that.

andrewsmith (approved):
This looks good to me!

core/rust/src/worker.rs
jkachmar: i think that doc is referring to how the `max_outstanding_{workflow,activity,nexus}_tasks` options are superseded by tuners for each task type (as opposed to the `max_concurrent_{workflow,activity,nexus}_task_polls` options, which have been superseded by a `PollerBehavior` for each task type).

since we control what we're handing to the Rust SDK we can construct a fixed-size tuner from the individual config fields (like how we construct `SimpleMaximum` pollers for the individual config fields we expose).

the only thing i'm not entirely sure about is why upstream decided to error out instead of implicitly falling back on constructing a fixed-size tuner... idk, might be worth asking them?

---

i think, longer term, we probably want to make some breaking changes to the API & drop the old fields in favor of some building tuner & poller structs from Haskell config records; i've tried to enumerate which fields should be going away in a future update [in this other branch](https://github.com/MercuryTechnologies/hs-temporal-sdk/compare/jkachmar/update-sdk-rust...jkachmar/workerconfig-documentation) open in #375 for reference.

PR #375, opened 2026-05-06
docs: core worker & client documentation

jkachmar:
## description

re-do of #348, this time with documentation for the client types as well.


azuzunaga (approved):
Nice!

PR #402, opened 2026-07-09
single-threaded, value-based workflow run machinery

jkachmar:
## description

this piggybacks off of #401: since we're using delimited continuations there, we can do away with the `Coroutine` machinery entirely, as well as removing the thread-based machinery for executing workflow runs.

now activation & suspension of workflow run tasks are managed in a similar manner to how fibers are (as delimited continuations) and execution state is tracked as a value in the `WorkflowInstance` record & advanced through `runActivation` under an `MVar` (which serializes access).

one interesting consequence here is that we don't need to do any STM bookkeeping or atomic modification of `IORef`s since all the workflow runs are single-threaded executions being stepped through.

locally there appear to be no performance impacts to this (E2E test suite took like 200 seconds before/after these changes).

this work fell out of trying to come up with a way to stub out the actual core backend & drive activations through a model-based test, and poking at ways to refactor the architecture to make _that_ goal easier.

not only does the approach here make that _significantly_ easier, but it also (i think) simplifies the execution model quite a bit.

#### `MVar`

the `MVar` is the only thing that feels a little uncomfortable, but we have masking discipline there to minimize the critical path & the worst case scenario is that something blows it up & the blast radius is limited to that workflow run i think? it might be worth a pathological test to make sure that throwing async exceptions at odd angles doesn't break anything

#### deadlock timeout

i added a config modifier to set deadlock timeout & then a test to validate it, and the test failed b/c one of the areas i exercised caused a deadlock in the core SDK; i have some ideas for what to do about that but it's still sorta an open note i guess.


jkachmar:
closed in favor of #414, which more accurately describes this as using delimited continuations for the workflow activation loop.

the single-threading fell out of this as a happy accident which simplifies the implementation.

PR #403, opened 2026-07-09
client: fix event streaming pagination bug

jkachmar:
## description

ran into a continue-as-new bug with the error: `RunId in the request does not match the NextPageToken.`

turns out this is the same bug as in #294, so we can resolve both & add a small regression test to confirm that this works as expected.

### DIFF
--- a/sdk/test/ClientHistorySpec.hs
+++ b/sdk/test/ClientHistorySpec.hs
@@ -1,0 +1,55 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module ClientHistorySpec where
+
+import Conduit (runConduit, sinkList, (.|))
+import Data.IORef
+import Data.ProtoLens (defMessage)
+import Lens.Family2 ((&), (.~), (^.))
+import Proto.Temporal.Api.History.V1.Message (HistoryEvent)
+import qualified Proto.Temporal.Api.History.V1.Message_Fields as History
+import Proto.Temporal.Api.Workflowservice.V1.RequestResponse
+  ( GetWorkflowExecutionHistoryRequest
+  , GetWorkflowExecutionHistoryResponse
+  )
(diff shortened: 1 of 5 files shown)


sdk/test/IntegrationSpec.hs
```diff
@@ -58,7 +58,6 @@ import IntegrationSpec.Updates
 import Lens.Family2
 import OpenTelemetry.Trace
 import qualified Proto.Temporal.Api.Failure.V1.Message_Fields as Failure
-import Proto.Temporal.Api.History.V1.Message (WorkflowExecutionFailedEventAttributes (..))
```
jkachmar, line 61: unused import

iand675 (approved):

sdk/test/ClientHistorySpec.hs
```diff
@@ -0,0 +1,55 @@
...
+          modifyIORef' reqsRef (<> [req])
+          readIORef respsRef >>= \case
+            (r : rest) -> writeIORef respsRef rest >> pure r
+            [] -> pure (historyResp [completedEvent] "") -- terminate if over-fetched
+    _ <- runConduit (streamEventsWith fetch FollowRuns req0 .| sinkList)
+    reqs <- readIORef reqsRef
+    -- Did not truncate on the non-terminal middle page (all three pages fetched).
+    length reqs `shouldBe` 3
+    -- The page after the continue-as-new targets run B, not the original run A.
+    ((reqs !! 2) ^. RR.execution . RR.runId) `shouldBe` "run-B"
```
jkachmar, line 55: the server uses a heuristic to determine how many will get put in a page, with the max requested being more a suggestion than anything else.

for example, if you `mapConcurrently` over 200 items and then `fetchHistory` with a 100 item limit you'll get a single page of 200 items.

sequentially generating a lot of history events makes the test long, so we just test the function directly and call it a day

PR #404, opened 2026-07-09
sdk: deadlock detection

jkachmar:
## description

this change adds a deadlock detection mechanism which checks for forward progress during workflow activation.

### DIFF
--- a/sdk/src/Temporal/Exception.hs
+++ b/sdk/src/Temporal/Exception.hs
@@ -305,7 +305,10 @@ data LogicBug = LogicBug LogicBugType
   deriving stock (Show)
 
 
-instance Exception LogicBug
+instance Exception LogicBug where
+  displayException (LogicBug WorkflowActivationDeadlock) =
+    "[TMPRL1101] Potential deadlock detected: workflow task did not yield to the runtime within the deadlock timeout"
+  displayException lb = show lb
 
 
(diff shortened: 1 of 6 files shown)


sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -893,8 +894,33 @@ newTrackedIVar = do
...
+the task, never influence command output.
+-}
+checkActivationDeadline :: InstanceM ()
+checkActivationDeadline = do
+  inst <- ask
+  liftIO $ do
+    n <- (+ 1) <$> readIORef inst.workflowFuel
+    writeIORef inst.workflowFuel n
+    -- Sample the clock every 1024th op.
+    when (n `rem` 1024 == 0) $
```
iand675: assuming this is a bit of an arbitrary check

sdk/src/Temporal/Exception.hs
```diff
@@ -305,7 +305,10 @@ data LogicBug = LogicBug LogicBugType
   deriving stock (Show)
 
 
-instance Exception LogicBug
+instance Exception LogicBug where
+  displayException (LogicBug WorkflowActivationDeadlock) =
+    "[TMPRL1101] Potential deadlock detected: workflow task did not yield to the runtime within the deadlock timeout"
```
jkachmar, line 310: this is apparently the error code we're supposed to use for this

PR #406, opened 2026-07-09
contrib: bounded concurrency utils

jkachmar:
## description

this is a long overdue extraction of some bounded concurrency utilities we've had in our production codebase for well over a year now.

this change is based on the non-cancelling `biselectOpt` branch because the changes that the delimited continuations feature branch that ran cancellations on losing branches of `biselectOpt` introduced considerable overhead for the sorts of workloads that these functions introduced.

### DIFF
--- a/sdk/src/Temporal/Contrib/BoundedConcurrency.hs
+++ b/sdk/src/Temporal/Contrib/BoundedConcurrency.hs
@@ -1,0 +1,467 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DeriveFoldable #-}
+{-# LANGUAGE DeriveFunctor #-}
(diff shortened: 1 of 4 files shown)


jkachmar:
i think this is going to be reworked a bit to base the concurrency primitives on top of an “awaitable” abstraction which only admits pollers (& the like) so the continuation we capture & retry on “losing” sides of the selection operation cannot include anything that might restart meaningful (command issuing) work like workflows or activities themselves.

jkachmar:
closed in favor of #422; the non-cancelling `biselectOptNoCancel` combinator preserved the API but introduced a small risk of rescheduling commands that would perform additional work.

instead, we opt to build off of an `Inflight` concurrency primitive that builds bounded concurrency directly off of sub-fiber scheduling with delimited continuations.

PR #407, opened 2026-07-09
workflow: provide non-cancelling `biselectOpt`

jkachmar:
## description

so it turns out that the delimited continuations work introduced a performance regression in one very specific area: `biselectOpt` used to construct `select`/`poll`-style concurrency primitives, which we had used in our production codebase to construct bounded concurrency helper functions (see #406, which is dependent on this changeset).

fixing the regression involved adding a `Disposition` type that selects between cancelling or dropping the losing branches of `biselectOpt`; `race` & `biselect` cancel, `poll` & `select`-likes drop.

as a result we recover most of the performance[^1] that came from cancelling the losing `biselectOpt` branches (some marginal performance overhead from switching to delimited continuations appears to be unavoidable).

[^1]: 150s -> 9s to 10s for the combined `unfoldConcurrentlyBounded` & `handleActionResult` tests in #406; before the delimited continuations work it was ~7s to ~8s

### DIFF
--- a/sdk/src/Temporal/Workflow.hs
+++ b/sdk/src/Temporal/Workflow.hs
@@ -175,6 +175,7 @@ module Temporal.Workflow (
   independently,
   biselect,
   biselectOpt,
+  biselectOptNoCancel,
   yield,
 
   -- * Interacting with running Workflows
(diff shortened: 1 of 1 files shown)


jkachmar:
small update: after talking with Ian about this i think there’s a bit of semantic drift between what the original behavior was & what the new behavior is with delimited continuations such that this probably isn’t the right approach.

instead, i’m going to try & provide a `select` primitive that’s explicit about taking “awaitable” types rather than `Workflow`s; that will limit us to only operating over types whose behavior is just polling for something & which can be retried on failure without the risk that they’ll produce new commands.

jkachmar:
closed in favor of https://github.com/MercuryTechnologies/hs-temporal-sdk/pull/420; adding a cancellation `Disposition` to `biselectOpt` was additional complexity & it produced a combinator that was somewhat unsafe with respect to scheduling additional commands.

instead of reusing it for concurrency exploration, we define a separate structure representing `Inflight` concurrency and use that to build bounded concurrency utils.

PR #416, opened 2026-07-12
Replace proto-lens with wireform-proto

iand675:
Generate Temporal protobuf bindings through the local wireform plugin and migrate SDK consumers to wireform's presence-aware generated records.

Why: proto-lens is slow and the developer experience of using lenses for it is bad for humans, Haddocks, LSPs, and LLMs.


core/src/Temporal/Core/Client.hs
```diff
@@ -358,7 +358,7 @@ type PrimRpcCall = Ptr CoreClient -> Ptr CRpcCall -> TokioCall CRPCError (CArray
 --
 -- This function is async-exception-safe: all Rust allocations are properly
 -- freed even if an async exception occurs during processing.
-call :: forall svc t. (HasMethodImpl svc t) => PrimRpcCall -> Client -> MethodInput svc t -> IO (Either RpcError (MethodOutput svc t))
+call :: forall req res. (MessageEncode req, MessageDecode res) => PrimRpcCall -> Client -> req -> IO (Either RpcError res)
```
jkachmar: is it worth having a `MessageCodec` constraint that encapsulates both en/decode constraints?

sdk/src/Temporal/Activity/Worker.hs
```diff
@@ -98,28 +100,37 @@ execute worker = runActivityWorker worker go
...
-      , workflowId = WorkflowId $ msg ^. AT.workflowExecution . P.workflowId
-      , runId = RunId $ msg ^. AT.workflowExecution . P.runId
-      , activityId = ActivityId $ msg ^. AT.activityId
-      , activityType = msg ^. AT.activityType
+      { workflowNamespace = Namespace $ fromMaybe "" msg.workflowNamespace
+      , workflowType = WorkflowType $ fromMaybe "" msg.workflowType
+      , workflowId = WorkflowId $ fromMaybe "" workflowExecution.workflowId
+      , runId = RunId $ fromMaybe "" workflowExecution.runId
+      , activityId = ActivityId $ fromMaybe "" msg.activityId
+      , activityType = fromMaybe "" msg.activityType
```
jkachmar: i think these are all `fold`

sdk/src/Temporal/Client/Schedule.hs
```diff
@@ -486,15 +518,15 @@ data ScheduleInfo = ScheduleInfo
...
-    , invalidScheduleError = p ^. S.invalidScheduleError
+    { actionCount = fromMaybe 0 (getField @"actionCount" p)
+    , missedCatchupWindow = fromMaybe 0 (getField @"missedCatchupWindow" p)
+    , overlapSkipped = fromMaybe 0 (getField @"overlapSkipped" p)
+    , runningWorkflows = V.toList (workflowExecutionFromProto <$> getField @"runningWorkflows" p)
+    , recentActions = V.toList (scheduleActionResultFromProto <$> getField @"recentActions" p)
+    , futureActionTimes = V.toList (timespecFromTimestamp <$> getField @"futureActionTimes" p)
+    , createTime = timespecFromTimestamp <$> getField @"createTime" p
+    , updateTime = timespecFromTimestamp <$> getField @"updateTime" p
+    , invalidScheduleError = fromMaybe "" (getField @"invalidScheduleError" p)
```
jkachmar: do we need `getField` or can these be use record dot accessors?

sdk/src/Temporal/Client/Schedule.hs
```diff
@@ -676,45 +720,48 @@ mkScheduleAction (workflowRef -> KnownWorkflow codec wfName) (WorkflowId wfId) o
...
-          & W.maybe'workflowExecutionTimeout .~ (durationToProto <$> opts.timeouts.executionTimeout)
-          & W.maybe'workflowRunTimeout .~ (durationToProto <$> opts.timeouts.runTimeout)
-          & W.maybe'workflowTaskTimeout .~ (durationToProto <$> opts.timeouts.taskTimeout)
-          & W.maybe'retryPolicy .~ (retryPolicyToProto <$> opts.retryPolicy)
-          & W.memo .~ convertToProtoMemo opts.memo
-          & W.searchAttributes .~ (defMessage & C.indexedFields .~ searchAttrs)
-          & W.header .~ (defMessage & C.fields .~ fmap convertToProtoPayload opts.headers)
+        mempty
+          { W.workflowId = Just wfId
+          , W.workflowType = Just ((mempty :: C.WorkflowType) {C.name = Just wfName})
```
jkachmar, line 725: i would have thought `WorkflowType` inferred here

sdk/src/Temporal/Client/Schedule.hs
```diff
@@ -856,13 +897,13 @@ patchSchedule c (ScheduleId s) p = liftIO $ do
...
-          & WF.identity .~ c.identity
-          -- TODO
-          & WF.requestId .~ p.requestId
+      ( mempty
+          { WF.namespace = Just (rawNamespace c.scheduleClientNamespace)
+          , WF.scheduleId = Just s
+          , WF.patch = Just (schedulePatchToProto p)
+          , WF.identity = Just c.identity
+          , WF.requestId = Just p.requestId
+          }
```
jkachmar, line 906: any idea what the todo was? this looks pretty much the same

sdk/src/Temporal/Nexus/Worker.hs
```diff
@@ -121,13 +118,14 @@ completeTask core token status = do
...
-        & NexusMsg.failure .~ (defMessage & NexusMsg.message .~ msg)
-    )
+  Nexus.NexusTaskCompletion'Status'Error $
+    NexusMsg.HandlerError
+      (Just "NOT_FOUND")
+      (Just (NexusMsg.Failure (Just msg) mempty Nothing []))
+      ( Just
+          ( if retryable
+              then NexusEnum.NexusHandlerErrorRetryBehavior'NexusHandlerErrorRetryBehaviorRetryable
+              else NexusEnum.NexusHandlerErrorRetryBehavior'NexusHandlerErrorRetryBehaviorNonRetryable
```
jkachmar, line 128: oof, some of these are kinda rough; proto lens had the same issue though so i guess it's win some, lose some?

sdk/src/Temporal/Client/Schedule.hs
iand675: Overloaded record fields require this sometimes, unfortunately.

sdk/src/Temporal/Client/Schedule.hs
iand675: TODO was to make sure we filled in all the fields, IIRC.

sdk/src/Temporal/Nexus/Worker.hs
iand675: yeah the fact that protobufs can have nested enums makes for some unpleasantness.

jkachmar (commented):
approving the content but as discussed out of band i think we might want to come up with a mechanism to have this code live alongside the proto-lens code so a consumer can quickly switch back if there was some weird edge case that shows up after an extended deployment and we need a way to switch back.

jkachmar (approved):

PR #417, opened 2026-07-12
sdk: speed up test suite

jkachmar:
## description

* tighten up threadDelay calls
* remove duplicate heartbeat tests
* fix QueryWorkflow RPC call to include queryId on the failing branch, which results in substantially shorter round-trip time (30s -> <100ms)

### DIFF
--- a/sdk/src/Temporal/WorkflowInstance.hs
+++ b/sdk/src/Temporal/WorkflowInstance.hs
@@ -474,6 +474,7 @@ applyQueryWorkflow queryWorkflow = do
       -- TODO, more useful error message
       pure $
         defMessage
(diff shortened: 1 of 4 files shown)


sdk/src/Temporal/WorkflowInstance.hs
```diff
@@ -474,6 +474,7 @@ applyQueryWorkflow queryWorkflow = do
       -- TODO, more useful error message
       pure $
         defMessage
+          & Command.queryId .~ rawQueryId baseInput.handleQueryId
```
iand675, line 477: Nice, this is why I want to get off of `proto-lens`, lol.

iand675 (approved):

PR #418, opened 2026-07-12
sdk: introduce 'Pending' and 'Awaitable'

jkachmar:
## notes

i’m not thrilled with the `awaitAny` formulation, i want to revisit it before i call this ready for review. #415 on the other hand is ready to go so maybe i’ll reorder these commits so that can merge first and this one after.

## description

this PR introduces the `Pending` type which wraps up a single `IVar` & its decode function, as well as an `Awaitable` typeclass which abstracts over `Pending`.

the goal here is to provide a primitive that explicitly only captures in-flight operations which do not admit cancellation and do not modify the command stream.

### DIFF
--- a/sdk/src/Temporal/Workflow.hs
+++ b/sdk/src/Temporal/Workflow.hs
@@ -130,6 +130,12 @@ module Temporal.Workflow (
   ChildWorkflowHandle,
   Wait (..),
   Cancel (..),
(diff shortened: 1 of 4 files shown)


jkachmar:
this ended up not being as clean as i'd hoped: `Task` doesn't quite fit this mold and `ChildWorkflowHandle` would require deeper restructuring. i'm going to try a different approach

PR #424, opened 2026-07-21
Fix activity worker finalization

jkachmar:
## description

When we cancel activities, we need to provide the Rust SDK enough time to /send/ cancellation signals to the Temporal server and receive a response.

If we use the same graceful shutdown period in both the Rust SDK & for our own wrapping wait period, then they will both fire near-simultaneously & it's possible that not all activities will be cancelled properly.

Instead, let's do the following:
* do NOT broadcast activity thread cancelation, let that be driven by the Rust side of things, so that it can send a message to the server & then trigger the exception that's propagated to the Haskell threads
* wrap `waitWorker` in a timeout that provides a 1 second buffer around the graceful shutdown period; this gives the Rust SDK enough time to issue RPCs to the Temporal Server after the graceful shutdown period has elapsed
* wrap the `finalizeShutdown` call in a 5s timeout to ensure that it doesn't block finalization if it hangs for some reason; this should give it enough time to issue RPCs (if it needs to) and if that's blocked for some reason then something's jammed up that we can't reasonably wait for if we're blocking the caller's finalization

### DIFF
--- a/sdk/src/Temporal/Activity/Worker.hs
+++ b/sdk/src/Temporal/Activity/Worker.hs
@@ -36,7 +36,6 @@ import Temporal.Exception
 import Temporal.Interceptor
 import Temporal.Payload
 import UnliftIO
-import UnliftIO.Concurrent (threadDelay)
 
 
 data ActivityWorker env = ActivityWorker
@@ -50,16 +49,6 @@ data ActivityWorker env = ActivityWorker
   , activityErrorConverters :: {-# UNPACK #-} ![ApplicationFailureHandler]
   , payloadProcessor :: {-# UNPACK #-} !PayloadProcessor
   }
-
-
-notifyShutdown :: MonadUnliftIO m => ActivityWorker env -> m ()
-notifyShutdown worker = do
-  let shutdownGracePeriod = fromIntegral (Core.gracefulShutdownPeriodMillis $ Core.getWorkerConfig worker.workerCore)
-  -- TODO logging here
-  when (shutdownGracePeriod > 0) $ do
-    threadDelay (shutdownGracePeriod * 1000)
-  running <- liftIO $ ListT.toList $ StmMap.listTNonAtomic $ worker.runningActivities
-  forConcurrently_ running $ \thread -> cancelWith (snd thread) WorkerShutdown
 
 
 newtype ActivityWorkerM env m a = ActivityWorkerM {unActivityWorkerM :: ReaderT (ActivityWorker env) m a}
--- a/sdk/src/Temporal/Worker.hs
+++ b/sdk/src/Temporal/Worker.hs
@@ -104,6 +104,7 @@ module Temporal.Worker (
   Core.TunerConfig (..),
   Core.SlotSupplierConfig (..),
   Core.ResourceBasedTunerConfig (..),
+
   -- ** Custom slot supplier
   Core.CustomSlotSupplier (..),
   Core.SlotReservationContext (..),
@@ -122,6 +123,7 @@ import Control.Monad.Catch
 import Control.Monad.Logger
 import Control.Monad.Reader
 import Control.Monad.State
+import qualified Data.ByteString as BS
 import Data.Either (lefts)
 import Data.Foldable
 import Data.HashMap.Strict (HashMap)
@@ -146,21 +148,20 @@ import qualified StmContainers.Map as StmMap
 import System.IO.Unsafe
 import Temporal.Activity.Definition
 import qualified Temporal.Activity.Worker as Activity
-import qualified Temporal.Nexus.Worker as Nexus
 import Temporal.Common
 import Temporal.Common.Async
-import Temporal.Workflow.Types (NexusClient (..), makeNexusClient)
 import qualified Temporal.Common.Logging as Logging
 import Temporal.Core.Client
-import qualified Data.ByteString as BS
 import Temporal.Core.Worker (InactiveForReplay)
 import qualified Temporal.Core.Worker as Core
 import Temporal.Exception
 import Temporal.Interceptor
+import qualified Temporal.Nexus.Worker as Nexus
 import Temporal.Payload (PayloadProcessor (..))
 import Temporal.Runtime
 import Temporal.Worker.Types
 import Temporal.Workflow.Definition
+import Temporal.Workflow.Types (NexusClient (..), makeNexusClient)
 import qualified Temporal.Workflow.Worker as Workflow
 import UnliftIO
 
@@ -377,11 +378,13 @@ setMaxOutstandingNexusTasks n = modifyCore $ \conf ->
     { Core.maxOutstandingNexusTasks = Just n
     }
 
+
 unsetMaxOutstandingNexusTasks :: ConfigM actEnv ()
 unsetMaxOutstandingNexusTasks = modifyCore $ \conf ->
   conf
     { Core.maxOutstandingNexusTasks = Nothing
     }
+
 
 setMaxConcurrentWorkflowTaskPolls :: Word64 -> ConfigM actEnv ()
 setMaxConcurrentWorkflowTaskPolls n = modifyCore $ \conf ->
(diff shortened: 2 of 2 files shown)


sdk/src/Temporal/Worker.hs
```diff
@@ -944,25 +949,43 @@ shutdown :: (MonadUnliftIO m) => Temporal.Worker.Worker actEnv -> m ()
...
+            timeout bufferShutdownMicros $ do
+              () <- case workerType of
+                Core.SReal -> Activity.forceShutdown workerActivityWorker
+                Core.SReplay -> pure ()
+              waitWorker worker
+
+  -- Throw a 'RuntimeError' exception if we don't finalize after a few seconds;
+  -- this should never happen in practice but it's worth keeping to ensure
+  -- that 'shutdown' finishes promptly if it's used as a finalizer.
+  let finalizeTimeoutMicros = 5_000_000
```
jkachmar, line 979: rule of thumb i used for these was give it enough breathing room to make a few RPCs if the server was backed up.

even then this feels overly generous so i'm open to pushback.

in the future we may want to make this configurable.

turnage (approved):

sdk/src/Temporal/Activity/Worker.hs
```diff
@@ -52,14 +51,18 @@ data ActivityWorker env = ActivityWorker
...
+return.
+
+__NOTE__: This is a backstop to guarantee that the worker promptly shuts down
+in the event that the Rust core SDK does not return control from /its/ shutdown
+sequence in a timely manner.
+-}
+forceShutdown :: MonadUnliftIO m => ActivityWorker env -> m ()
+forceShutdown worker = do
+  running <- liftIO $ ListT.toList $ StmMap.listTNonAtomic worker.runningActivities
+  forConcurrently_ running $ \(_token, thread) -> cancelWith thread WorkerShutdown
```
parsonsmatt: This is potentially thread wasteful, though I'm unsure if the public api of async allows for a more native approach - [`cancelMany`](https://www.stackage.org/haddock/lts-24.51/async-2.2.6/src/Control.Concurrent.Async.Internal.html#cancelMany) is implemented as `mapM (\(Async _ t) -> throwTo t AsyncCancelled)` followed by `mapM waitCatch`. Not concurrent, but potentially more prompt and require less scheduling work.

sdk/src/Temporal/Worker.hs
```diff
@@ -944,25 +949,43 @@ shutdown :: (MonadUnliftIO m) => Temporal.Worker.Worker actEnv -> m ()
...
-  err' <- OT.inSpan workerTracer "finalizeShutdown" defaultSpanArguments $ liftIO $ Core.finalizeShutdown workerCore
-  case err' of
-    Left err -> throwIO err
-    Right () -> pure ()
+  -- Add 1s of buffer time to the graceful shutdown timeout so the Rust SDK
+  -- has time to return when/if it hits the graceful shutdown period; without
+  -- the buffer, our timeout fires a bit earlier than Rust's does and we can
+  -- end up killing the process too soon.
+  let bufferShutdownMicros = 1_000_000
+      gracefulShutdownMicros = bufferShutdownMicros + 1000 * fromIntegral (Core.gracefulShutdownPeriodMillis $ Core.getWorkerConfig workerCore)
```
parsonsmatt: One second is a lot. How long does it take Rust to do this?

parsonsmatt: Huh, the default here is 0 https://github.com/MercuryTechnologies/hs-temporal-sdk/blob/103d4ae6f8ef44411051b6398ff97af9d6a837d3/core/src/Temporal/Core/Worker.hs#L616

So the typical flow *was* to immediately send shutdown notifications. Workers would get `notifyShutdown`, and then immediately we'd `waitWorker` to proceed, and then promptly go into `finalizeShutdown`.

sdk/src/Temporal/Worker.hs
```diff
@@ -944,25 +949,43 @@ shutdown :: (MonadUnliftIO m) => Temporal.Worker.Worker actEnv -> m ()
...
+
+  drained <-
+    OT.inSpan workerTracer "waitWorker" defaultSpanArguments $
+      restore $
+        timeout gracefulShutdownMicros (waitWorker worker)
```
parsonsmatt: Now, instead of waiting indefinitely for the worker to acknowledge shutdown, we have a `timeout`.

Potentially `waitWorkerSTM`? I think I saw that in the API along with `registerDelay` - may be more economicaly

sdk/src/Temporal/Worker.hs
```diff
@@ -944,25 +949,43 @@ shutdown :: (MonadUnliftIO m) => Temporal.Worker.Worker actEnv -> m ()
...
+    Just () -> pure ()
+    Nothing ->
+      void $
+        OT.inSpan workerTracer "waitWorker (backstop)" defaultSpanArguments $
+          restore $
+            timeout bufferShutdownMicros $ do
+              () <- case workerType of
+                Core.SReal -> Activity.forceShutdown workerActivityWorker
+                Core.SReplay -> pure ()
+              waitWorker worker
```
parsonsmatt: This `Nothing` case means that we reached our timeout. By default, that timeout is the 1 second magic number above, plus however many milliseconds have been configured for graceful shutdown. Now we do another timeout, this time *forcing* the worker to shutdown. We assume this takes at most one second.

sdk/src/Temporal/Worker.hs
parsonsmatt: Five seconds is an eternity in shutdown logic, agree. do we have any metrics on how long this may end up requiring?

sdk/src/Temporal/Worker.hs
parsonsmatt: ok I think that we should not be adding a second to whatever shutdown period has been configured without updating the docs to say "actually whatever value you put in will  have one second added to it." 

Which feels a little strange? if they've specified a graceful shutdown, then in prior versions of the code, after calling `gracefulShutdown`, they'll get X milliseconds of wait time before everything shuts down. But now there is no such delay- the notice to shutdown is delivered promptly, we just have a *timeout* on how long that delay takes now.

parsonsmatt (commented):
I'm running a cold so not at 100%. but I think this is doing two things: getting rid of the graceful shutdown time (possibly accidentally?) and adding some timeouts + forcing the shutdown process so it completes more promptly.

sdk/src/Temporal/Worker.hs
jkachmar: we don’t, and yeah i agree it is an eternity.

the main issue is that each of these stages potentially issues RPCs to the server so we have whatever round trip time that is + delays for retries. the primary thing we want here is to afford these cleanup actions some time to run.

in practice none of these timeouts should hold shutdown open for the full amount of time; i think the second `waitWorker` after we broadcast cancelation might not be worth it.

a lot of these are last ditch efforts at giving the worker a chance to clean up if it gets SIGTERM’d by ECS

sdk/src/Temporal/Activity/Worker.hs
jkachmar: okay yeah i think this actually doesn't even make sense anymore; if we issue the cancelation request through the Rust SDK and it fails to arrive at the activity in time then i think everything should just get torn down without any grace period.

sdk/src/Temporal/Worker.hs
jkachmar: so the issue here is that we pass the grace period to the Rust SDK which bounds _its_ cleanup actions by that amount; if we use the same value for both _and_ we hit the timeout period then the Haskell side will finish first and start gracelessly tearing things down.

so what's happening now is we give the Rust SDK 1 second of overhead to make any RPCs it might need to (with whatever that overhead might be) before getting back to us.

and because we are no longer throwing the async exception ourselves, the Rust SDK is responsible for respecting the graceful shutdown period before issuing its cancelation requests to the server (which delivers them back to us).

sdk/src/Temporal/Worker.hs
jkachmar: yeah i think this isn't strictly necessary and we can just drop it; if we've failed to shut things down in the initial block of work then all `forceShutdown` would have done is ensure that the async exception hits the `Activity` functions & runs their finalizers...

sdk/src/Temporal/Worker.hs
jkachmar: also, for what it's worth, the original function was blocking so a bounded 5s timeout is still an improvement

turnage (approved):

PR #425, opened 2026-07-21
Unmask activity execution

jkachmar:
## description

`unmask` activity execution and swap from an `MVar` as a sync point around activity registration to a `TVar` which handles de-registration if the activity completes before we even progress to that point.

this avoids the concern about missing registration (and thus not being able to cancel a running activity) if something funny happens in the middle of a run.

### DIFF
--- a/sdk/src/Temporal/Activity/Worker.hs
+++ b/sdk/src/Temporal/Activity/Worker.hs
@@ -165,7 +165,6 @@ requireActivityNotRunning tt m = do
     Nothing -> m
 
 
--- TODO, where should async exception masking happen?
 applyActivityTaskStart :: (MonadUnliftIO m, MonadLogger m) => AT.ActivityTask -> TaskToken -> AT.Start -> ActivityWorkerM actEnv m ()
 applyActivityTaskStart _tsk tt msg = do
   w <- ask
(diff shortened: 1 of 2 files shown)


sdk/src/Temporal/Activity/Worker.hs
```diff
@@ -210,20 +209,19 @@ applyActivityTaskStart _tsk tt msg = do
...
+        -- Unmask activity execution, but catch /all/ exceptions so we can
+        (completion :: Either SomeException (Either String Payload)) <- unmask $ liftIO $ UnliftIO.trySyncOrAsync $
           w.activityInboundInterceptors.executeActivity env input $ \env' input' -> do
             case HashMap.lookup info.activityType w.definitions of
               Nothing -> throwIO $ RuntimeError ("Activity type not found: " <> T.unpack info.activityType)
               Just ActivityDefinition {..} ->
                 runReaderT (unActivity $ activityRun input') (actEnv env')
-                  `finally` (takeMVar syncPoint *> atomically (StmMap.delete tt w.runningActivities))
-        completionMsg <- case ef >>= first (toException . ValueError) of
+                  `finally` atomically (writeTVar finished True *> StmMap.delete tt w.runningActivities)
```
jkachmar: this avoids the MVar critical path entirely, i think?

sdk/src/Temporal/Activity/Worker.hs
```diff
@@ -165,7 +165,6 @@ requireActivityNotRunning tt m = do
     Nothing -> m
 
 
--- TODO, where should async exception masking happen?
```
jkachmar, line 168: well hey at least we can answer that question now

sdk/src/Temporal/Activity/Worker.hs
```diff
@@ -214,15 +213,16 @@ applyActivityTaskStart _tsk tt msg = do
     -- before we start running it. This is important because we need to be able to cancel
     -- it later if the orchestrator requests it.
     mask_ $ do
-      syncPoint <- newEmptyMVar
-      runningActivity <- asyncLabelled (T.unpack $ T.concat ["temporal/worker/activity/start/", Core.namespace c, "/", Core.taskQueue c]) $ do
-        (ef :: Either SomeException (Either String Payload)) <- liftIO $ UnliftIO.trySyncOrAsync $ do
+      finished <- newTVarIO False
+      runningActivity <- asyncLabelledWithUnmask (T.unpack $ T.concat ["temporal/worker/activity/start/", Core.namespace c, "/", Core.taskQueue c]) $ \unmask -> do
+        -- Unmask activity execution, but catch /all/ exceptions so we can
```
turnage: so we can ...

turnage (approved):

andrewsmith (commented):
This generally looks good to me from what little I know about exception masking.

Is there any way we can have an automated test for this?

sdk/src/Temporal/Activity/Worker.hs
```diff
@@ -214,15 +213,16 @@ applyActivityTaskStart _tsk tt msg = do
...
+        (ef :: Either SomeException (Either String Payload)) <- unmask $ liftIO $ UnliftIO.trySyncOrAsync $
```
parsonsmatt: I believe this `unmask` location is safe. An async exception could be delivered during the `HashMap.lookup` , but once you have the `ActivityDefinition`, the `finally (runReaderT ...) (atomically ...)` will re-mask and ensure that cleanup is called.

parsonsmatt (approved):

PR #427, opened 2026-08-06
Fix query handler startup race

iand675:
## Summary
- make query dispatch wait for a workflow’s first synchronous evaluation to finish
- prevent an immediate query after `Client.start` from observing a top-level handler as missing
- add a 100-query immediate-start regression

## Why
`Client.start` confirms persistence before the worker has necessarily registered its synchronous query handlers. The temporal-outbox worker’s immediate `getConfig` query could therefore fail with `QueryNotFound`.

## Verification
`cabal test temporal-sdk-tests --test-option=--match --test-option=Query --test-show-details=direct`

## Risk
A workflow that blocks in unsafe IO before its first yield also blocks startup queries; normal workflow code must not do that.


iand675:
Closing: this PR was accidentally opened from a stale branch. The replacement contains only the startup readiness fix.

PR #428, opened 2026-08-06
Fix query handler startup race

iand675:
## Summary
- make query dispatch wait for a workflow’s first synchronous evaluation to finish
- prevent an immediate query after `Client.start` from observing a top-level handler as missing
- add a query immediate-start regression test

## Why
`Client.start` confirms persistence before the worker has necessarily registered its synchronous query handlers. The temporal-outbox worker’s immediate `getConfig` query could therefore fail with `QueryNotFound`.

## Verification
`cabal test temporal-sdk-tests --test-option=--match --test-option=Query --test-show-details=direct`

### DIFF
--- a/sdk/src/Temporal/Contrib/OpenTelemetry.hs
+++ b/sdk/src/Temporal/Contrib/OpenTelemetry.hs
@@ -19,7 +19,7 @@ the interceptor to your Temporal client and worker configuration.
 module Temporal.Contrib.OpenTelemetry where
 
 import Control.Monad.Catch
-import Control.Monad (forM_)
+import Control.Monad (forM_, void)
 import Control.Monad.IO.Class
 import Data.IORef
 import qualified Data.HashMap.Strict as HashMap
@@ -34,7 +34,7 @@ import GHC.IO (unsafePerformIO)
 import OpenTelemetry.Propagator.W3CBaggage (decodeBaggage, encodeBaggage) 
 import OpenTelemetry.Context (Context)
 import qualified OpenTelemetry.Context as Ctxt
-import OpenTelemetry.Context.ThreadLocal (attachContext, getContext)
+import OpenTelemetry.Context.ThreadLocal (attachContext, detachContext, getContext)
 import OpenTelemetry.Propagator
 import OpenTelemetry.Propagator.W3CTraceContext (decodeSpanContext, encodeSpanContext)
 import OpenTelemetry.Trace.Core
@@ -236,10 +236,13 @@ makeOpenTelemetryInterceptor = do
                                 ]
                         }
                 span <- createSpan tracer ctxt ("RunWorkflow:" <> rawWorkflowType input.executeWorkflowInputType) spanArgs
-                _ <- attachContext $ Ctxt.insertSpan span ctxt
+                priorContext <- attachContext $ Ctxt.insertSpan span ctxt
                 atomicModifyIORef' workflowSpans $ \spans ->
                   (HashMap.insert (rawRunId input.executeWorkflowInputInfo.runId) span spans, ())
-                next input
+                let restoreContext = case priorContext of
+                      Nothing -> void detachContext
+                      Just prior -> void $ attachContext prior
+                next input `finally` restoreContext
             , finalizeWorkflow = finalizeWorkflow
             , handleQuery = \input next -> do
                 -- Only trace this if there is a header, and make that span the parent.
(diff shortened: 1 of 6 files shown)


jkachmar:
yeah i think in an ideal world we wouldn't need to install these barriers but i can't think of a better option either.

jkachmar (approved):

sdk/src/Temporal/Contrib/OpenTelemetry.hs
```diff
@@ -149,22 +151,35 @@ tracerKey = unsafePerformIO Vault.newKey
 --    */
 -- CONTINUE_AS_NEW = 'ContinueAsNew',
 
+makeOpenTelemetryInterceptor :: IO (Interceptors env)
+
 -- TODO, we will need to account for replays when we support them
-makeOpenTelemetryInterceptor :: MonadIO m => m (Interceptors env)
 makeOpenTelemetryInterceptor = do
   tracerProvider <- getGlobalTracerProvider
+  workflowSpans <- liftIO $ newIORef mempty
```
jkachmar: if we're in `IO` now we can drop the lift

iand675: hm, on second thought I'll stick with `makeOpenTelemetryInterceptor` stay generic

sdk/src/Temporal/Contrib/OpenTelemetry.hs
```diff
@@ -97,11 +92,34 @@ headersPropagator =
...
+    , injector = \c hs -> case Ctxt.lookupSpan c of
+        Nothing -> pure hs
+        Just s -> do
+          (traceParentHeader, traceStateHeader) <- encodeSpanContext s
+          pure
+            . Map.insert "traceparent" (Payload traceParentHeader mempty)
+            . Map.insert "tracestate" (Payload traceStateHeader mempty)
+            $ hs
+    }
+#endif
```
jkachmar, line 116: is this just duplicated or what?

sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -339,15 +340,20 @@ newWorkflowGenM g = Workflow $ \_ -> Done . WorkflowGenM <$> newIORef g
...
-#if MIN_VERSION_random(1,3,0)
+
 instance ThawedGen StdGen Workflow where
-#endif
   thawGen = newWorkflowGenM
+#else
+instance FrozenGen StdGen Workflow where
+  type MutableGen StdGen Workflow = WorkflowGenM
+  freezeGen g = Workflow $ \_ -> Done <$> readIORef (unWorkflowGenM g)
+  thawGen = newWorkflowGenM
```
jkachmar, line 355: what's going on here?

sdk/src/Temporal/Contrib/OpenTelemetry.hs
iand675: fourmolu gets really weird about formatting when CPP is in the mix, and it turns out that duplicating the code makes it not crash...

rgoomar: Seems so from what I see too

sdk/src/Temporal/Workflow/Internal/Monad.hs
iand675: same fourmolu issue as the other spot

sdk/src/Temporal/Workflow/Worker.hs
```diff
@@ -156,18 +144,20 @@ handleActivation activation = inSpan' "handleActivation" (defaultSpanArguments {
...
-                && not (isInitActivation && isJust (job ^. Activation.maybe'signalWorkflow))
-            withoutStart = filter keepJob (activation ^. Activation.jobs)
-        case withoutStart of
-          [] -> pure ()
-          otherJobs -> atomically $ writeTQueue inst.activationChannel (activation & Activation.jobs .~ otherJobs)
+      forM_ mInst $ \(inst, handledInitialJobs) -> do
+        unless handledInitialJobs do
+          -- Signals in the initialization activation were already buffered into the
+          -- instance (see applyStartWorkflow), so drop them here rather than
+          -- delivering them a second time.
```
jkachmar, line 151: is it worth refactoring the signal handler stuff to use this mechanism as well (for consistency) or do you think these should necessarily be handled differently?

jkachmar (approved):

sdk/src/Temporal/Workflow/Worker.hs
iand675: probably, I'm going to defer that idea for a separate PR in order to keep shipping the outbox worker.

PR #432, opened 2026-08-17
Fix workflow activation head-of-line blocking

iand675:
## Problem

The startup-query fix in `d622f395` moved `handleActivation` from a linked async per activation into the sole `WorkflowWorker.execute` poll loop. If one workflow blocks during startup or activation handling, that worker stops polling every other workflow assigned to it. Temporal has already leased those tasks, so unrelated workflows can hit their workflow-task `StartToClose` timeout.

This change restores the required separation: the global loop only polls and dispatches. Workflow code runs outside it.

## Execution model

There are two ordering requirements:

1. Different workflow runs must not block one another.
2. Activations for the same run must enter the workflow instance in poll order.

The second requirement is subtle because activation B can be polled before activation A has finished creating and publishing its `WorkflowInstance`. Looking up the instance is therefore not enough to serialize A and B.

### Per-run activation tails

`workerActivationTails` is a map from run ID to the newest reservation for that run. A reservation contains a lifecycle ticket and a `TMVar ()` completion signal.

When the poll loop receives an activation, one STM transaction:

1. reserves a lifecycle ticket;
2. reads the current tail's completion signal as this activation's predecessor;
3. installs this activation's completion signal as the new tail.

The loop then forks the handler. The handler waits for its predecessor, handles the activation, and fills its own completion signal in a masked `finally`.

```mermaid
sequenceDiagram
    participant P as Poll loop
    participant T as Tail map for run R
    participant A as Handler A
    participant B as Handler B
    participant W as Workflow executor R

    P->>T: reserve A, set tail(R) to doneA
    P->>A: fork
    P->>T: reserve B, predecessor is doneA, set tail(R) to doneB
    P->>B: fork
    A->>W: publish or enqueue activation A
    A->>T: fill doneA
    T-->>B: predecessor released
    B->>W: enqueue activation B
    B->>T: fill doneB
```

Only the newest reservation stays in the map. Each successor temporarily retains its direct predecessor signal, so the chain is transitive without storing an unbounded list in the map. Older links become collectible as the chain advances. A cache-removal activation removes the map entry only if it is still the newest tail; it cannot erase a later reservation.

The tails order dispatch, not workflow evaluation. Each `WorkflowInstance` owns one long-lived executor that consumes its activation channel. This keeps a run serial while allowing different run IDs to proceed concurrently.

## Shutdown and eviction

The poll loop's admission lifecycle is now a pure `ActivationLoop` state machine:

- `Polling`: new handlers may reserve tickets.
- `Draining`: admission is closed permanently.
- The active ticket set is exactly the handlers admitted but not yet finalized.
- Shutdown proceeds only in `Draining` with an empty active set.

Reservation happens before fork, and completion stays in the handler's masked `finally`. Shutdown first enters `Draining`, waits for all admitted handlers, then cancels workflow executors. Eviction marks an instance cancelled, cancels and joins its executor, and only then finalizes interceptor resources.

## Adding a bit of formalism to ensure safety

This PR factors out the pure state transition model so that we can test it some without spinning up a worker.

This is not a machine-checked proof of the entire concurrent worker, but it takes some gentle steps in that direction with some executable checks under test.

### 1. Production and tests share the transition functions

`reserveActivation`, `completeActivation`, and `beginDraining` are pure functions. The production STM code calls those functions directly rather than reimplementing their rules. Tests therefore exercise the same admission and completion transitions used by the worker.

### 2. Property tests check the concrete model behaviour.

Hedgehog generates long sequences containing reservations, active completions, duplicate completions, stale and never-issued tickets, repeated drains, and both serializations of the reserve/drain race. It checks that:

- tickets are not reused while active;
- completion removes exactly its own ticket;
- stale or duplicate completion cannot retire newer work;
- draining never reopens admission;
- shutdown cannot report drained while an admitted handler is missing its finalizer.

This is broad bug search over concrete ticket values, but random testing alone is not exhaustive.

### 3. Collapsing an infinite search space into a finite quotient

Our worker model's relevant state has two dimensions:

 1. Is the worker still accepting activations?
   - Polling
   - Draining
  
 2. How many activation handlers are still active?
   - 0
   - 1
   - 2+

 That produces six buckets:

| Worker phase | Active handlers |
|--------------|----------------| 
| Polling      | 0               |
| Polling      | 1               |
| Polling      | 2+              |
| Draining     | 0               |
| Draining     | 1               |
| Draining     | 2+              |

 We then start from Polling / 0 and try every action that matters:

 - reserve a new activation;
 - finish an activation;
 - begin draining;
 - try an invalid or stale completion.

 For every resulting bucket, we try all the actions again. We continue until every action only leads to a bucket we have already checked.

 That is what “exhausted” means here: the exploration reached a fixed point. There are no unexplored buckets or transitions left in this simplified
 model.

 ### Why 2+ is enough

 The shutdown property does not care exactly how many handlers there are. It cares whether all handlers have finished.

 From 2+, completing one handler can leave:

 - exactly 1; or
 - still 2+.

 The model checks both possibilities. Therefore 3, 50, and 10,000 active handlers do not add a new kind of behavior for this property—they just repeat
 the 2+ → 2+ transition before eventually reaching 2+ → 1.

 We still keep the concrete Hedgehog model with real ticket identities to check details such as:

 - the same ticket cannot complete twice;
 - an unknown ticket cannot complete;
 - finishing ticket A cannot remove ticket B;
 - ticket identities are not reused during a worker’s lifetime.

 The six-state abstraction is specifically for the lifecycle question:

 Can the worker claim to be drained while an admitted activation handler is still running, or admit new work after draining begins?

 ### What it does not prove

 It does not exhaust every possible production execution. It abstracts away:

 - actual thread scheduling;
 - exceptions between arbitrary IO operations;
 - the per-run tail map;
 - workflow execution behavior;
 - STM implementation mistakes that diverge from the pure model.

### 4. Worker-level regressions exercise the original failure mode

The integration tests use the real worker to check that:

- a workflow blocked during startup does not starve an unrelated workflow;
- 200 concurrent startup queries plus 100 signals complete within a bound;
- the event-subscriber parent/child workflow shape completes without workflow-task timeout.

## Other lifecycle fixes

- Workflow execution context is restored around each resumed activation without retaining thread-local guards across suspension.
- Query, update, and activity OpenTelemetry contexts restore the previous thread context on success or failure.
- Completed workflows keep servicing queries until eviction.
- Poll shutdown performs one drain/cancellation pass.

## Verification

- `cabal test temporal-sdk-tests --test-option=--match --test-option='activation loop lifecycle' --test-show-details=direct` - 9 examples, 0 failures
- `cabal test temporal-sdk-tests --test-option=--match --test-option='Query' --test-show-details=direct` - 19 examples, 0 failures, 1 pending
- `cabal test temporal-sdk-tests --test-show-details=direct` - 522 examples, 0 failures, 66 pending

Toward: GRC-130

### DIFF
--- a/sdk/src/Temporal/Contrib/OpenTelemetry.hs
+++ b/sdk/src/Temporal/Contrib/OpenTelemetry.hs
@@ -149,6 +149,8 @@ headersBaggagePropagator =
     }
 #endif
 
+
+restoreContext :: Maybe Context -> IO ()
 #if MIN_VERSION_hs_opentelemetry_api(1,0,0)
 restoreContext = void . detachContext
 #else
@@ -312,12 +314,14 @@ makeOpenTelemetryInterceptor = do
                         --   ]
                         }
                 ctxt <- extract headersPropagator input.handleQueryInputHeaders Ctxt.empty
-                _ <- attachContext ctxt
-                case Ctxt.lookupSpan ctxt of
-                  Nothing -> next input
-                  Just _ ->
-                    inSpan'' tracer ("HandleQuery:" <> input.handleQueryInputType) spanArgs $ \_ -> do
-                      next input
+                priorContext <- attachContext ctxt
+                ( case Ctxt.lookupSpan ctxt of
+                    Nothing -> next input
+                    Just _ ->
+                      inSpan'' tracer ("HandleQuery:" <> input.handleQueryInputType) spanArgs $ \_ -> do
+                        next input
+                  )
+                  `finally` restoreContext priorContext
             , handleUpdate = \input next -> do
                 -- Only trace this if there is a header, and make that span the parent.
                 -- We do not put anything that happens in an update handler on the workflow
@@ -331,22 +335,21 @@ makeOpenTelemetryInterceptor = do
                         , attributes = mempty
                         }
                 ctxt <- performUnsafeNonDeterministicIO $ extract headersPropagator input.handleUpdateInputHeaders Ctxt.empty
-                _ <- performUnsafeNonDeterministicIO $ attachContext ctxt
-                case Ctxt.lookupSpan ctxt of
+                priorContext <- performUnsafeNonDeterministicIO $ attachContext ctxt
+                result <- try @_ @SomeException $ case Ctxt.lookupSpan ctxt of
                   Nothing -> next input
                   Just _ -> do
                     span <- performUnsafeNonDeterministicIO $ createSpan tracer ctxt ("HandleUpdate:" <> input.handleUpdateInputType) spanArgs
-                    result <- try $ next input
-                    case result of
+                    updateResult <- try @_ @SomeException $ next input
+                    performUnsafeNonDeterministicIO $ case updateResult of
                       Left err -> do
-                        performUnsafeNonDeterministicIO do
-                          setStatus span (Error $ T.pack $ show err)
-                          recordException span mempty Nothing err
-                          endSpan span Nothing
-                        throwM (err :: SomeException)
-                      Right res -> do
-                        performUnsafeNonDeterministicIO $ endSpan span Nothing
-                        pure res
+                        setStatus span (Error $ T.pack $ show err)
+                        recordException span mempty Nothing err
+                        endSpan span Nothing
+                      Right _ -> endSpan span Nothing
+                    either throwM pure updateResult
+                performUnsafeNonDeterministicIO $ restoreContext priorContext
+                either throwM pure result
             , validateUpdate = \input next -> do
                 let spanArgs =
                       defaultSpanArguments
@@ -354,12 +357,14 @@ makeOpenTelemetryInterceptor = do
                         , attributes = mempty
                         }
                 ctxt <- extract headersPropagator input.handleUpdateInputHeaders Ctxt.empty
-                _ <- attachContext ctxt
-                case Ctxt.lookupSpan ctxt of
-                  Nothing -> next input
-                  Just _ ->
-                    inSpan'' tracer ("ValidateUpdate:" <> input.handleUpdateInputType) spanArgs $ \_ -> do
-                      next input
+                priorContext <- attachContext ctxt
+                ( case Ctxt.lookupSpan ctxt of
+                    Nothing -> next input
+                    Just _ ->
+                      inSpan'' tracer ("ValidateUpdate:" <> input.handleUpdateInputType) spanArgs $ \_ -> do
+                        next input
+                  )
+                  `finally` restoreContext priorContext
             }
       , workflowOutboundInterceptors =
           WorkflowOutboundInterceptor
@@ -420,9 +425,9 @@ makeOpenTelemetryInterceptor = do
                         }
 
                 ctxt <- extract headersPropagator input.activityHeaders Ctxt.empty
-                _ <- attachContext ctxt
-                inSpan'' tracer ("RunActivity:" <> input.activityInfo.activityType) spanArgs $ \_span -> do
-                  next env input
+                priorContext <- attachContext ctxt
+                inSpan'' tracer ("RunActivity:" <> input.activityInfo.activityType) spanArgs (\_span -> next env input)
+                  `finally` restoreContext priorContext
             }
       , activityOutboundInterceptors = ActivityOutboundInterceptor
       , clientInterceptors =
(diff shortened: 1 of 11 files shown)


sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -783,21 +776,24 @@ data WorkflowInstance = WorkflowInstance
   , workflowCommands :: {-# UNPACK #-} !(TVar (Reversed WorkflowCommand))
   , workflowSequenceMaps :: {-# UNPACK #-} !(TVar SequenceMaps)
   , workflowSignalHandlers :: {-# UNPACK #-} !(IORef (HashMap (Maybe Text) (Vector Payload -> Workflow ())))
-  , workflowBufferedSignals :: {-# UNPACK #-} !(IORef (HashMap Text (Endo [Vector Payload])))
-  -- ^ Signals that arrived before their handler was registered.
-  -- These are buffered and delivered when setSignalHandler is called.
-  -- Uses Endo for O(1) append (diff-list style).
+  , -- | Signals that arrived before their handler was registered.
```
iand675, line 779: fourmolu change

sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -520,10 +514,9 @@ newIVar = do
   pure IVar {ivarId = 0, ..}
 
 
-{- | Allocate an IVar with a unique id drawn from the per-instance counter.
-Tracked IVars have their blocking call stacks recorded so that built-in
-queries can report all concurrent stacks.
--}
+-- | Allocate an IVar with a unique id drawn from the per-instance counter.
```
iand675, line 517: fourmolu change

sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -340,20 +339,15 @@ newWorkflowGenM g = Workflow $ \_ -> Done . WorkflowGenM <$> newIORef g
...
   applyRandomGenM = applyWorkflowGen
 
-#if MIN_VERSION_random(1,3,0)
+
 instance FrozenGen StdGen Workflow where
   type MutableGen StdGen Workflow = WorkflowGenM
   freezeGen g = Workflow $ \_ -> Done <$> readIORef (unWorkflowGenM g)
 
-
+#if MIN_VERSION_random(1,3,0)
```
iand675, line 347: tweaking the placement of these lets fourmolu work on the file despite it using CPP

sdk/src/Temporal/Workflow/Worker.hs
```diff
@@ -50,11 +72,14 @@ data EvictionWithRunID = EvictionWithRunID
...
 
-data WorkflowWorker = forall ty.
+data WorkflowWorker
+  = forall ty.
   Core.KnownWorkerType ty =>
   WorkflowWorker
   { workerWorkflowFunctions :: {-# UNPACK #-} !(HashMap Text WorkflowDefinition)
   , runningWorkflows :: {-# UNPACK #-} !(StmMap.Map RunId WorkflowInstance)
+  , workerActivationLoop :: {-# UNPACK #-} !(TVar ActivationLoop.ActivationLoop)
+  , workerActivationTails :: {-# UNPACK #-} !(StmMap.Map RunId (ActivationLoop.ActivationTicket, TMVar ()))
```
iand675, line 82: See the PR description explaining why this change was introduced

sdk/src/Temporal/Contrib/OpenTelemetry.hs
```diff
@@ -312,12 +314,14 @@ makeOpenTelemetryInterceptor = do
...
-                    inSpan'' tracer ("HandleQuery:" <> input.handleQueryInputType) spanArgs $ \_ -> do
-                      next input
+                priorContext <- attachContext ctxt
+                ( case Ctxt.lookupSpan ctxt of
+                    Nothing -> next input
+                    Just _ ->
+                      inSpan'' tracer ("HandleQuery:" <> input.handleQueryInputType) spanArgs $ \_ -> do
+                        next input
+                  )
+                  `finally` restoreContext priorContext
```
iand675, line 324: here and elsewhere I don't think it's a dramatic issue that we weren't restoring context previously, but should make sure that issue attribution is more accurate in the future

azuzunaga (approved):
Answered my own q, looks good. Thanks for the tests.

PR #434, opened 2026-08-19
bridge: spawn Tokio futures instead of blocking the calling thread

jkeuhlen:
## Summary

- Tokio-backed FFI calls now return after scheduling Rust work. Haskell waits on an interruptible MVar.
- If the Haskell wait is interrupted, the Rust task continues. A reaper keeps result storage alive and frees it after completion.
- Worker finalization waits for Core shutdown before unwrapping the worker.
- Worker and ephemeral server shutdown are single-flight. Repeated calls share the first result and never reuse a consumed Rust pointer.
- Public worker shutdown runs on an unmasked thread so timeouts work inside cleanup handlers.

## Validation

- Rust unit tests and Clippy
- GHC 9.10.1 build
- Full SDK suite: 513 examples, 0 failures, 66 pending


core/src/Temporal/Core/Worker.hs
```diff
@@ -92,10 +93,23 @@ import Temporal.Core.CTypes
...
 
 
+{- | Temporarily identify a synchronous worker FFI call in thread dumps.
+
+GHC cannot deliver an asynchronous exception while one of these calls is in
+progress. Replacing the label for the duration of the call makes the blocked
+operation visible through 'GHC.Conc.Sync.threadLabel'.
+-}
+withFfiThreadLabel :: String -> IO a -> IO a
+withFfiThreadLabel lbl act = mask $ \restore -> do
```
iand675, line 106: handy!

jkeuhlen: Yeah it had to add this in order to figure out wtf was going on iiuc lol.

core/src/Temporal/Core/Worker.hs
```diff
@@ -92,10 +93,23 @@ import Temporal.Core.CTypes
...
+  tid <- Control.Concurrent.myThreadId
+  prev <- GHC.Conc.Sync.threadLabel tid
```
iand675, line 108: `threadLabe`l was introduced in base-4.18, so we need to bump the `base` constraints in cabal

core/src/Temporal/Core/Worker.hs
```diff
@@ -137,7 +166,11 @@ data Worker (ty :: WorkerType) = Worker
 
 withWorker :: forall ty a. KnownWorkerType ty => Worker ty -> (Ptr (Worker ty) -> IO a) -> IO a
 withWorker w f = withRuntime w.workerRuntime $ \_ -> do
-  ptr <- readIORef w.workerPtr
+  ptr <-
```
iand675, line 169: Before this PR, `Core.validateWorker` blocked inside the foreign call. This PR makes it return from FFI and wait interruptibly on the callback MVar at
 core/src/Temporal/Internal/FFI.hs:150.

 If a timeout or `killThread` now interrupts validation:

 1. `Core.newWorker` has already returned an owned worker.
 2. The Tokio validation task continues.
 3. The new reaper correctly frees the validation result and slot allocations.
 4. Nothing closes `workerCore`.
 5. `startWorker` never returns, so an outer bracket (`startWorker` ...) shutdown has no Worker value to release.

The outer bracket does not protect this. Acquisition runs MaskedInterruptible, and a blocking `readMVar` remains an async-exception delivery point under that state.

This can strand the Core worker and any runtime ownership retained through its outstanding task. The same internal cleanup boundary would also fix the existing leak when validation returns Left.


So, I think to fix this, we'd need to do something like:

Make the portion of startWorker after Core.newWorker exception-safe. Ownership should transfer to the returned high-level Worker only after initialization succeeds. Until then, an onException or inner bracketOnError must run the correct pre-start Core cleanup sequence. Audit the later loop-spawning portion under the same rule so cancellation after one loop starts cannot leave a partially initialized worker.

core/src/Temporal/Core/Worker.hs
```diff
@@ -846,13 +898,35 @@ This should be called only after 'initiateShutdown' has resolved and/or both pol
...
+      WorkerClosed -> throwIO WorkerAlreadyClosed
+
+  when shouldStart $ do
+    spawned <-
+      try $
+        forkIO $ do
+          outcome <-
+            try $
+              withRuntime w.workerRuntime $ \_ ->
+                withTokioAsyncCall
```
iand675, line 918: `withTokioAsyncCallWithAbandon` seems to exist specifically because a successfully started server needs `shutdownAbandonedEphemeralServer`, not the normal result destructor. I think we should see if we can get some direct test coverage around cancelling startup after the Rust future is scheduled, then prove the eventual process and port are cleaned up.

sdk/test/ReplaySpec.hs
```diff
@@ -19,7 +26,89 @@ import TestHelpers
 
 
 spec :: Spec
-spec = withTestServer_ tests
+spec = do
+  describe "Tokio FFI interruption" $ do
+    specify "interrupts a blocked poll and reaps its eventual result" $
```
iand675, line 33: OK, so I think this proves interruption and successful subsequent shutdown. It doesn't directly prove the returned Rust allocation was reclaimed. A drop counter in a bridge test fixture could help us make that explicit so we're confident that the rust side is doing the right thing too.

iand675 (approved):

core/rust/src/runtime.rs
```diff
@@ -164,56 +212,173 @@ pub struct HsCallback<A, E> {
...
+    /// The C ABI entry point must return after scheduling. Haskell then waits on
+    /// an interruptible `takeMVar`; using `block_on` here would instead keep it
+    /// inside the foreign call until the future completed, preventing
+    /// `timeout` and `killThread` from interrupting the wait.
     pub fn future_result_into_hs<F, T, E>(&self, callback: HsCallback<T, E>, fut: F)
     where
         F: Future<Output = Result<T, E>> + Send + 'static,
-        T: RawPointerConverter<T>,
-        E: RawPointerConverter<E>,
+        T: RawPointerConverter<T> + 'static,
```
jkachmar, line 271: i guess there’s no way to thread lifetime evidence here so it has to be `'static` same as the `Future`?

PR #435, opened 2026-08-25
Fix workflow instance executor lifecycle

iand675:
## Problem

Workflow instance creation forked an executor before registration completed. A setup exception could strand a blocked thread, and a losing duplicate publication could start a detached executor for the same run. Activation dispatch also attached OTel context without detaching it.

## Changes

- Fork the executor only after built-in query registration and successful run-ID publication.
- Start only the insertion winner and label executor threads with their run ID.
- Restore the activation thread's OTel context with `finally detachContext`.
- Add a `signalWithStart` regression that requires exactly one workflow-body execution.
- Remove GHC 9.6 and 9.8 from the Nix development, package, coverage, and overlay matrix. GHC 9.10 is now the sole supported Nix target.

## Verification

- `cabal test temporal-sdk-tests --test-options='--match signalWithStart'`
- `cabal test temporal-sdk-tests --test-options='--match __stack_trace'`
- `nix build .#hs-temporal-suite-ghc910 --no-link`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,7 +51,7 @@
           haskellUtils = import ./nix/utils/haskell.nix pkgs;
           inherit (import ./nix/utils/matrix.nix) ghcVersions;
 
-          # Generate coverage packages for each GHC version
+          # Generate coverage packages for supported GHC versions
           coveragePackages = builtins.listToAttrs (
             builtins.map (ghcVersion: {
               name = "coverage-${ghcVersion}";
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -94,18 +94,6 @@
         haskell-development = final: prev: {
           haskell = (prev.haskell or { }) // {
             packages = (prev.haskell.packages or { }) // {
-              ghc96 = prev.haskell.packages.ghc96.extend (
-                prev.lib.composeManyExtensions [
-                  (self.haskellOverlays.dependencies.default final)
-                  (self.haskellOverlays.hs-temporal-sdk final)
-                ]
-              );
-              ghc98 = prev.haskell.packages.ghc98.extend (
```
iand675, line 103: Jake is working on a parallel PR that requires GHC >= 9.10, so go ahead and drop these from CI.

nix/utils/matrix.nix
```diff
@@ -6,8 +6,6 @@
   ];
 
   ghcVersions = [
-    "ghc96"
-    "ghc98"
     "ghc910"
```
jkachmar, line 9: note to future self: add ghc912 to the matrix & do some nixpkgs cleanup…

jkeuhlen (approved):
LGTM. I think this will solve the hang problem I encountered.

PR #439, opened 2026-08-27
perf: log full activity results at Debug instead of Info

jkeuhlen:
## Summary

`Temporal.Workflow`'s activity `waitAction` logged the **entire `ResolveActivity` proto — including the full result payload bytes — via `show` at Info level** on every activity completion:

```haskell
Logging.logInfo ("Activity result: " <> Text.pack (show res))
```

At the typical Info log level this unconditionally allocates a String/Text proportional to the payload size for every completed activity in every workflow, and dumps payload contents into production logs. It was also the only `logInfo` in the module — every comparable trace statement in the SDK uses `logDebug`.

This changes it to `logDebug`. With monad-logger, filtered debug messages are dropped before the message thunk is forced, so the `show` never runs unless debug logging is enabled.

Incidental changes required to get the file past pre-commit hooks (it had drifted):
- fourmolu formatting fixes (import ordering, doc-comment style on the Nexus section)
- `{-# HLINT ignore "Use ?~" #-}` pragma, matching the existing suppression in `Temporal.Client.Schedule` (lens-family doesn't provide `?~`)

## Test plan

- `cabal build temporal-sdk:lib:temporal-sdk` passes (nix dev shell, GHC 9.10)
- All pre-commit hooks (fourmolu, hlint, hpack, etc.) pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sdk/src/Temporal/Workflow.hs
+++ b/sdk/src/Temporal/Workflow.hs
@@ -1,3 +1,4 @@
+{-# HLINT ignore "Use ?~" #-}
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DataKinds #-}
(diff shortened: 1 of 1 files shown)


sdk/src/Temporal/Workflow.hs
```diff
@@ -448,7 +449,7 @@ startActivityFromPayloads (KnownActivity codec name) opts typedPayloads = ilift
       Task
         { waitAction = do
             res <- getIVar resultSlot
-            Logging.logInfo ("Activity result: " <> Text.pack (show res))
+            Logging.logDebug ("Activity result: " <> Text.pack (show res))
```
jkeuhlen, line 452: Seems like we could just make this `DEBUG` instead of `INFO` in general to reduce log outputs. 

(Other changes here are from the precommit hooks)

jkachmar (approved):

PR #440, opened 2026-08-28
Fix workflow executor lifecycle boundaries

iand675:
## Problem

The workflow executor conflated workflow-task failure, workflow completion, and executor eviction:

- A `WorkflowActivationDeadlock` raised by the SDK watchdog was converted into a terminal workflow failure instead of failing the activation, so Temporal could not evict and replay the workflow task.
- Workflow interceptor finalization was delayed behind the post-completion query loop until cache eviction. Clients could observe a completed workflow while its `RunWorkflow` span was still open and therefore absent from exported spans.
- Escaped inbound-interceptor failures and later eviction could attempt finalization more than once.

## Change

- Finalize escaped non-async inbound-interceptor failures with their workflow failure result.
- Elect one finalization per workflow instance atomically so completion or failure wins over eviction.
- Let SDK `LogicBug` failures escape top-level workflow conversion and reach activation-failure handling.
- Finalize workflow interceptors before publishing a terminal command, then keep the executor alive for post-completion queries.
- Add integration regressions for deadline eviction and replay, exactly-once finalization, activity tracing, and continue-as-new tracing.

## Prior failure

With the old finalization ordering, the deterministic activity regression failed with:

```text
workflow result was observable before RunWorkflow finalization: Right 1
1 example, 1 failure
```

The continue-as-new regression also observed only one of the expected two `RunWorkflow` spans.

## Risk

SDK invariant failures now fail the workflow activation rather than terminating the workflow execution. Workflow finalizers run before terminal command publication instead of waiting for cache eviction. Post-completion query handling remains active, and eviction remains an idempotent fallback.

## Verification

- `direnv exec . cabal test temporal-sdk --test-options='--match "retries a resumed activation that exceeds the deadlock timeout"'`
- `direnv exec . cabal test temporal-sdk --test-options='--match "Workflow finalization"'`
- `direnv exec . cabal test temporal-sdk --test-options='--match "OpenTelemetry workflow finalization"'`
- `direnv exec . cabal test temporal-sdk --test-options='--match "query on completed workflow"'`

The final OpenTelemetry suite passes with 2 examples and 0 failures.

### DIFF
--- a/sdk/src/Temporal/Workflow/Internal/Monad.hs
+++ b/sdk/src/Temporal/Workflow/Internal/Monad.hs
@@ -790,6 +790,8 @@ data WorkflowInstance = WorkflowInstance
   , workflowCancellationVar :: {-# UNPACK #-} !(IVar ())
   , workflowDeadlockTimeout :: Maybe Int
   , workflowVault :: {-# UNPACK #-} !(IORef Vault)
+  , -- | Ensures inbound interceptor finalizers observe one terminal outcome.
+    workflowFinalized :: {-# UNPACK #-} !(IORef Bool)
   , -- These are how the instance gets its work done
     activationChannel :: {-# UNPACK #-} !(TQueue Core.WorkflowActivation)
   , executionThread :: {-# UNPACK #-} !(IORef (Async ()))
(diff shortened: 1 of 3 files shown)


sdk/src/Temporal/Workflow/Internal/Monad.hs
```diff
@@ -954,7 +956,9 @@ data WorkflowInboundInterceptor = WorkflowInboundInterceptor
        . ExecuteWorkflowInput
       -> (ExecuteWorkflowInput -> IO a)
       -> IO a
-  , finalizeWorkflow
+  , -- | Called once when the workflow completes, fails, or is evicted before a
+    -- terminal outcome is known. A 'Nothing' result denotes that eviction.
```
jkeuhlen, line 960: I might be reading this wrong, but this function doesn't seem like it has a result type, isnt' it `IO ()`? Where is the result this comment is mentioning?

jkeuhlen (approved):

PR #441, opened 2026-09-02
Expose priority for activities and child workflows

tylerholien:
## Problem

Activities and child workflows could not override priority, and `signalWithStart` dropped it.

## Changes

- Add priority options for activities and child workflows.
- Propagate priority through Core commands and `signalWithStart`.
- Add integration coverage.

## Verification

- `cabal test temporal-sdk-tests --test-options='--match priority'`


jkachmar (approved):
this looks good to me; the Priority proto is defined in their common module anyway so it makes sense we share this between client/workflow modules via our own `Common.hs`.

everything else is either a test or a mechanical move so it's a lot less churn than it looks like. thanks for the contribution!

PR #443, opened 2026-09-10
core: make client connection state explicit

jkachmar:
## description

* Expose connection readiness and retry connections on failed client initialization
* Wake any client handles awaiting an active connection if the connection is closed, to prevent clients from waiting indefinitely
* Clone core client & runtime handles instead of borrowing raw pointers, this prevents a potential use-after-free when `closeClient` is called while another caller is waiting for or actively using the client

### notes

basically `ClientRef` gets to be `Clone` and then we introduce `Temporal.Core.Client.Internal.Connection` to explicitly model the connection lifecycle.

this is a bit of ceremony, but i think it's worthwhile to:

- avoid borrowing mutable references across FFI calls when acquiring client & runtime handles; they should be cloned whenever we pass a handle back to some caller in Haskell [^1]
- explicitly model the state of a `Client` handle, in particular the lifecycle of the connection established to a Temporal server through the FFI call during the client's backgrounded initialization [^2]
- propagate the outcome of a connection attempt to all callers that may be waiting on it [^3]

the explicit connection state modeling could be avoided if we didn't want to handle connection negotiation in the background, but that was added explicitly by e8e0f159 in 2023 so i'm assuming it was done in service of something useful.

[^1]: i _think_ the only way we could share a mutable reference is if we explicitly locked around areas where we need to express that the Haskell code maintains exclusive access to the reference. with `clone()` the lock only needs to be held long enough to make a copy, and with `Arc<SharedClient>` clones are cheap

[^2]: as best as i can tell we want to explicitly model this so we can talk about whether the underlying handle to Rust code can be safely accessed

[^3]: this is probably something that doesn't pop up much in a long-running production deployment, but it _does_ tend to happen in a test suite that's cycling through clients very quickly.


sdk/test/IntegrationSpec/ClientLifecycle.hs
```diff
@@ -0,0 +1,208 @@
...
+        Core.ClientConnectionError _ -> True
+        _ -> False
+      isClosed = \case
+        Core.ClientClosedError -> True
+        _ -> False
+
+  forM_ [False, True] $ \throughRpc ->
+    it ("recovers unavailable initialization through " <> if throughRpc then "an RPC" else "readiness") $ do
+      port <- Server.getFreePort
+      executable <- findExecutable "temporal" >>= maybe (throwIO $ userError "temporal executable missing") pure
```
jkachmar, line 70: maybe it's worth extracting the test helpers from `sdk` that handle this dev server lifecycle stuff

core/rust/src/client.rs
```diff
@@ -253,8 +259,10 @@ pub fn connect_client(
 
             match retry_client_result {
                 Ok(retry_client) => Ok(ClientRef {
-                    retry_client,
-                    runtime,
+                    client: Arc::new(SharedClient {
```
jkachmar: i think `inner` is more idiomatic here actually?

jkachmar: either way we want this to be `Arc` so clones are cheap

core/rust/temporal_bridge.h
```diff
@@ -1096,6 +1109,19 @@ const struct CArray_CArray_u8 *hs_temporal_runtime_fetch_logs(struct RuntimeRef
...
 void hs_temporal_runtime_free_logs(const struct CArray_CArray_u8 *logs);
 
+/**
+ * Clone a runtime handle, sharing the underlying runtime.
+ *
+ * The source remains owned by the caller. Release the returned handle exactly
+ * once with `hs_temporal_free_runtime`; either handle may outlive the other.
+ *
+ * # Safety
+ * `runtime` must be non-null, aligned, and valid for shared access throughout
```
iand675: Curious about the alignment requirement here. Seems like there's something for me to learn but dunno what it's about

core/rust/src/runtime.rs
```diff
@@ -440,3 +441,18 @@ pub unsafe extern "C" fn hs_temporal_runtime_free_logs(logs: *const CArray<CArra
...
+
+/// Clone a runtime handle, sharing the underlying runtime.
+///
+/// Release the returned handle exactly once with `hs_temporal_free_runtime`; either handle may outlive the other.
+///
+/// # Safety
+/// `runtime` must be non-null, aligned, and valid for shared access throughout
+/// this call.
+///
+/// The caller must prevent concurrent destruction or mutation of the source wrapper.
```
iand675: are we actually ensuring this

core/rust/temporal_bridge.h
```diff
@@ -209,6 +209,19 @@ void hs_temporal_drop_client(struct ClientRef *client);
  */
 void hs_temporal_drop_rpc_error(struct CRPCError *error);
 
+/**
+ * Clone a client handle, sharing its connection and runtime.
+ *
+ * Release the returned handle exactly once with `hs_temporal_drop_client`; either handle may outlive the other.
```
iand675, line 215: yeah this is pretty great

iand675: we could probably go back to using `ForeignPtr` in this case!

core/src/Temporal/Core/Client/Internal/Connection.hs
```diff
@@ -0,0 +1,206 @@
...
+
+notifyOutcome :: Outcome -> Either SomeException () -> IO ()
+notifyOutcome (Outcome outcome) result = void $ tryPutMVar outcome result
+
+
+-- | Check whether this connection attempt is still pending.
+isConnecting :: ClientHandle a -> Outcome -> IO Bool
+isConnecting conn outcome =
+  withConnState conn $
+    pure . \s -> case s of
```
iand675: nit: `pure . \case`

core/src/Temporal/Core/Client.hs
```diff
@@ -88,6 +90,14 @@ foreign import ccall "hs_temporal_connect_client" raw_connectClient :: Ptr Runti
 foreign import ccall "hs_temporal_drop_client" raw_freeClient :: Ptr CoreClient -> IO ()
 
 
+-- These unsafe imports clone handles synchronously; they do not wait for the
+-- network and they do not invoke Haskell callbacks.
+foreign import ccall unsafe "hs_temporal_clone_client" raw_cloneClient :: Ptr CoreClient -> IO (Ptr CoreClient)
+
+
+foreign import ccall unsafe "hs_temporal_clone_runtime" raw_cloneRuntime :: Ptr Runtime -> IO (Ptr Runtime)
```
iand675, line 98: Ok this part has always been a bit squirrelly to me, in pretty much all iterations of the SDK. Does clone of a client do the "right thing" with the runtime too?

core/rust/temporal_bridge.h
jkachmar: i think the alignment note was just some slop that got through; really all we require is that the handle we use here was produced by the Rust FFI and not modified by Haskell at all

core/rust/src/runtime.rs
jkachmar: for _this_ commit no, but the lifetime hazard here already existed (just with the borrowed mutable reference) i think? either way concurrent destruction & mutation are handled as part of the API changes that land in #444 & #445

core/src/Temporal/Core/Client.hs
jkachmar: so `raw_cloneClient` clones `ClientRef`, and after this change `ClientRef` is just an `Arc<SharedClient>` which retains _its_ runtime reference.

which is to say that after this commit cloning a client increments a ref count but does _not_ make a copy of the runtime, whereas `raw_cloneRuntime` is `Arc<CoreRuntime>` so _that_ clone will just increment _that_ ref count.

core/rust/temporal_bridge.h
jkachmar: i think so yea, although in that case it'd just be an extra guard right? like we'd still be managing the lifetime of the reference with all of this machinery & the `ForeignPtr` would be extra insurance that if a reference ever escaped we'd have a finalizer attached to it so GC could reclaim it rather than it being a leak forever.

PR #445, opened 2026-09-10
core: prevent Temporal runtime use-after-free

jkachmar:
## description

* Clone runtime handles under a lock before use
* Make repeated and concurrent destruction safe
* Ensure cloned runtimes are freed
* Generalize the client’s runtime ownership model for all runtime handles
* Release cloned handle locks after the call into tokio is made

### depends on

- #443
- #444


sdk/test/IntegrationSpec/ClientLifecycle.hs
```diff
@@ -220,16 +236,15 @@ spec = describe "Client lifecycle" $ do
             (runLoggingT (Core.connectClient runtime config) logger)
             Core.closeClient
             (\client -> takeMVar entered >> pure client)
-    bounded
-      ( bracket acquire Core.closeClient $ \client -> do
+        releaseLogger = void $ tryPutMVar proceed ()
```
iand675, line 239: don't really understand what this is going for

jkachmar: this is the same cleanup that
```
`finally` void (tryPutMVar proceed ())
```
was doing further down in the test just pulled out to a named value; maybe that's not helpful since you aren't really releasing the logger but trying to unblock the `MVar` that's used to synchronize the test?

